PR #59, opened 2026-05-13
fix(slippy): read-your-own-writes overlay closes async-insert aggregate race

pkuzmenko-mycarrier:
## Summary

Adds a read-your-own-writes overlay to the slip aggregate write path. Closes an async-insert race where `insertComponentState`'s ClickHouse INSERT was not visible to the immediately-subsequent `loadComponentStates` SELECT, causing `routing_slips.<step>_status` to be permanently stuck at `running` even after all components finished.

Confirmed in prod on slips:
- `d15f1808-5727-4df4-83d8-c526afd2c79e` (MC.Invoice)
- `51a2cb11-960b-4974-b131-2d85513cb4ba` (MC.Invoice)

## Root cause

`UpdateStepWithHistory` flow (`clickhouse_store.go:78`):

```
insertComponentState(...)                                      // async INSERT
  → updateAggregateStatusFromComponentStatesWithHistory(...)
       → s.Load(...) → hydrateSlip → loadComponentStates(...)  // SELECT misses fresh row
       → computeAggregateStatus(...) → returns stale "running"
       → s.Update(routing_slips, builds_status=running)        // wrong write
```

Both ops share a single `driver.Conn` with no transaction. ClickHouse `async_insert=1` means INSERT returns OK before the row is visible to subsequent SELECTs on the same connection (~200ms flush window). `computeAggregateStatus` reads pre-insert state, `Update()` writes it, version check passes (no concurrent writer) → no retry fixes it. ReplacingMergeTree eventually merges the row but `routing_slips` write-back never re-fires.

### Commit 1 `1981532` — Initial overlay

`overlayComponentState` — pure Go function. After `insertComponentState`, merges the just-written component into the in-memory `slip.Aggregates[aggregateStepName]` slice before `computeAggregateStatus`. Zero ClickHouse overhead, preserves async-insert throughput.

### Commit 2 `97c408f` — Round-1 review corrections (CRITICAL)

PR review round 1 found the initial fix was **mechanically incomplete**:

- **T1 (CRITICAL)**: `Update()` → `BuildStepColumnsAndValues` reads `slip.Steps[step].Status` for the scalar `{step}_status` Enum8 column. Overlay only patched `slip.Aggregates`, never `slip.Steps`. So `builds_status` would stay `running` despite correct aggregate JSON. **Fix:** after `overlayComponentState`, recompute aggregate from `slip.Aggregates[aggregateStepName]` and apply to `slip.Steps[aggregateStepName]` via `ApplyStatusTransition`. Applied in both `updateAggregateStatusFromComponentStates` and `updateAggregateStatusFromComponentStatesWithHistory`.

- **T2 (CRITICAL)**: New-component append path used raw `stepName` (could be alias `"build"`) instead of canonical aggregate key (`"builds"`). `BuildAggregateColumnsAndValues` only reads canonical key. **Fix:** `overlayComponentState` signature now accepts `aggregateStepName`. Both callers resolve and pass it.

- **T3 (IMPORTANT)**: Original regression test asserted only on `slip.Aggregates` slice, missed `slip.Steps` divergence. **Fix:** added `TestOverlayUpdatesStepsStatus_T1Regression` asserting `slip.Steps["builds"].Status == completed` post-overlay + `TestOverlayNewComponentUnderCanonicalKey_T2Regression` for T2 path.

- **T4, T5 (SUGGESTION)**: Documentation gaps on `latestComponentTime` zero-time semantics and `insertedAt` clock-skew safety on retry. **Fix:** doc comments added.

### Commit 3 (round-2 polish, on this branch)

PR review round 2 verified all round-1 fixes correct and complete. 3 new suggestions:

- **N1 (SUGGESTION)**: `overlayComponentState` doc-comment described an unreachable `""` fallback path that would have reintroduced T2 bug if a future caller hit it. **Fix:** removed misleading doc line, added defensive guard `if aggregateStepName == "" { return }` on the component-overlay path, simplified dead `bucketKey` fallback.

- **N2 (SUGGESTION)**: Recompute block passes all components (incl. pending placeholders) to `computeAggregateStatus`, while `applyComponentStatesToAggregate` filters to active-only. Same function returns different results depending on call site. Pre-existing divergence, neutral in current scenario. **Fix:** comment added above both recompute blocks documenting the divergence + tracking alignment as separate follow-up.

- **N3 (SUGGESTION)**: Skipped — regression test inline mirror is intentional, integration coverage exists.

## Spec alignment (STATE_MACHINE_V3.md)

| Invariant | Impact |
|---|---|
| I1 (aggregate from event log) | Preserved — overlay IS the just-appended event |
| Append-only events | Preserved |
| Terminal-success monotonic | Preserved |
| **I5** (cached column == event-log-derived) | **Strengthened** — eliminates permanent stale window. External CH readers retain bounded ~200ms staleness during async flush (unchanged); fix eliminates permanent staleness from write-back path. |
| Single authoritative write per terminal event | Preserved |

`.github/STATE_MACHINE_V3.md` updated with I5 overlay mechanism note. `.github/PROJECT_STATE.md` updated with full fix entry under "Recent Changes — May 13, 2026".

## Why this approach (Option 3 — in-memory overlay)

| Option | Approach | Trade-off |
|---|---|---|
| 1 — `SETTINGS wait_for_async_insert=1` | Sync flush before INSERT returns | Adds per-insert latency, drops async throughput |
| 2 — Retry/poll loop in `updateAggregateStatus*` | Re-query SELECT until row visible | Amplifies CH read load, stacks with CLI retries |
| 3 — Overlay (this PR) | Merge in Go memory | Zero CH overhead, deterministic, uses data already in hand |

Writer already constructs the row it just inserted — merging into in-memory Slip before `Update()` is free and deterministic.

## Cache layer note

`slippy-api/internal/infrastructure/cache.go` is **passthrough stub** today. `CachedSlipReader.{Load, LoadByCommit, FindByCommits, FindAllByCommits, InvalidateByCorrelationID}` all delegate without caching. Redis/Dragonfly client wired but GET/SET not implemented; Helm/K8s manifests not deployed; `DRAGONFLY_HOST` empty everywhere.

**This PR does not touch the cache layer.** When the cache activates (separate PR), write path's `invalidate` (called at `slippy-api/slip_write_handler.go:364`) must:

1. DELETE cached slip entry by correlation_id for `Load` keys
2. DELETE any cached `by-commit` keys that map to this correlation_id
3. DELETE cached `find-by-commits` / `find-all-by-commits` result sets containing this correlation_id

Without this, cache hits would return stale `routing_slips` values from before the overlay write. **Tracked as separate follow-up.**

## Test plan

```
TestOverlayComponentState_AddsNewComponent                       PASS
TestOverlayComponentState_ReplacesOlderEntry                     PASS
TestOverlayComponentState_LeavesNewerEntryAlone                  PASS
TestOverlayComponentState_OnlyTouchesNamedComponent              PASS
TestOverlayComponentState_NilSlip                                PASS
TestOverlayComponentState_PipelineLevelStep                      PASS
TestUpdateAggregateStatusFromComponentStates_AsyncInsertRace     PASS  ← original regression
TestOverlayUpdatesStepsStatus_T1Regression                       PASS  ← T1 regression
TestOverlayNewComponentUnderCanonicalKey_T2Regression            PASS  ← T2 regression
```

**Gates:**

| Gate | Result |
|---|---|
| `go build ./...` | PASS |
| `go vet ./...` | PASS |
| `gofmt -l .` | Clean |
| `go test ./... -race -count=1 -timeout=3m` | PASS (slippy 3.7s, slippytest 1.7s) |

**Coverage:** package overall 83.3%; `overlayComponentState` 96.7%; `latestComponentTime` 100%; `updateAggregateStatusFromComponentStates` 88.9%; `updateAggregateStatusFromComponentStatesWithHistory` 81.7%.

## Risk

**Low.** Scoped to two private functions. No schema change. No change to `insertComponentState`, ClickHouse settings, or `conflictVersion` retry path. Behavior change is strictly additive: newer overlay wins, older skipped, missing component added under canonical key.

External CH readers (HyperDX dashboards, ad-hoc queries) retain existing bounded async-flush staleness — unchanged. Fix only eliminates permanent stale state from write-back.

## Rollback

Revert. No DB migration. No infrastructure changes.

## Follow-ups (separate PRs)

1. **Cache invalidation** — when Dragonfly cache activates, implement `invalidate` for write path in `slippy-api/infrastructure/cache.go` (see Cache layer note).
2. **Gap B — crash recovery** — if slippy-api crashes between `insertComponentState` and routing_slips Update, no mechanism re-fires aggregate. WAL or periodic reconciler needed.
3. **Stuck-slip backfill** — `d15f1808`, `51a2cb11`, plus any others stuck before this fix lands need manual remediation. CollapsingMergeTree requires `-1` cancel + `+1` corrected pair.
4. **Placeholder-divergence alignment** — `computeAggregateStatus` filtering paths differ between `applyComponentStatesToAggregate` (active-only) and recompute block (all incl. placeholders). Neutral in current scenario; alignment worth doing.
5. **Component alias double-entry (cosmetic)** — closed by N1 guard; future callers must resolve aggregate key before invoking overlay.

## Review trail

- Round 1: 5 findings (2 critical, 1 important, 2 suggestions) — all resolved in commit `97c408f`
- Round 2: 3 suggestions — N1 + N2 applied in round-2 polish commit; N3 dismissed per devils-advocate verdict
- Security: 0 findings across both rounds (no auth, no SQL changes, no new imports, no secrets in test data)

### DIFF
--- a/.github/PROJECT_STATE.md
+++ b/.github/PROJECT_STATE.md
@@ -167,6 +167,50 @@ When edge cases are detected, warnings are logged with context. See [resolveAndA
 
 ## Recent Changes
 
+### May 13, 2026 — Async-Insert Race Fix in Aggregate Write-Back (PR #59)
+
+**Context:** Production slips `d15f1808-5727-4df4-83d8-c526afd2c79e` and `51a2cb11-960b-4974-b131-2d85513cb4ba` (both MC.Invoice) observed permanently stuck at `builds_status = running` even after all components completed. Root cause: ClickHouse `async_insert=1` server-side meant the row inserted by `insertComponentState` was not visible to the immediately-subsequent `loadComponentStates` SELECT on the same `driver.Conn`. `computeAggregateStatus` read pre-insert state and `Update()` wrote it back; the conflict-retry version check passed (no concurrent writer) so no retry fixed it. ReplacingMergeTree eventually merged the row but `routing_slips` write-back never re-fired. I5 invariant permanently violated.
+
+#### Fix (`slippy/clickhouse_store.go`)
+
+1. **`overlayComponentState`** — new pure function. Read-your-own-writes pattern: after `insertComponentState`, merge the just-written component into the in-memory `slip.Aggregates[aggregateStepName]` slice. Closes the visibility race in Go memory without changing ClickHouse settings (preserves async-insert throughput).
+2. **`latestComponentTime`** — helper for age comparison (max of `CompletedAt` / `StartedAt`).
+3. **Wiring** — both `UpdateStep` (~L449) and `UpdateStepWithHistory` (~L496) capture `insertedAt := time.Now()` before INSERT, thread `componentName`/`status`/`insertedAt` into the aggregate functions. Both `updateAggregateStatusFromComponentStates` (~L1836) and `updateAggregateStatusFromComponentStatesWithHistory` (~L1969) call `overlayComponentState` after `Load()` and before `Update()`.
+4. **Recompute block (review-driven correction)** — after the overlay, recompute aggregate status from the merged `slip.Aggregates[aggregateStepName]` and apply to `slip.Steps[aggregateStepName]` via `ApplyStatusTransition`. Required because `Update()` → `BuildStepColumnsAndValues` reads the scalar `{step}_status` Enum8 column from `slip.Steps`, not `slip.Aggregates`. Without this, the overlay patched the aggregate JSON column correctly but the scalar column stayed stale. **Identified by PR review round 1, finding T1.**
+5. **Canonical bucket key (review-driven correction)** — `overlayComponentState` now takes `aggregateStepName` as a parameter and uses it as the bucket key when appending a new component (rather than using the raw `stepName` which may be a component-type alias like `"build"`). `BuildAggregateColumnsAndValues` only reads the canonical key (e.g. `"builds"`). **Identified by PR review round 1, finding T2.**
+6. **Defensive guard + doc polish (round 2)** — `overlayComponentState` now early-returns when `aggregateStepName == ""` on the component-overlay path (pipeline-level sentinel with `componentName == ""` still uses `stepName` directly). Doc-comment rewritten to state the non-empty precondition explicitly. Dead `bucketKey` fallback removed. **Identified by PR review round 2, finding N1.**
+7. **Placeholder-divergence note (round 2)** — comment added above both recompute blocks explaining that they pass `slip.Aggregates[aggregateStepName]` (all components incl. pending placeholders) to `computeAggregateStatus`, vs. `applyComponentStatesToAggregate` which filters to active-only. Semantically neutral in the convergence path; full alignment tracked as separate follow-up. **Identified by PR review round 2, finding N2.**
+
+#### Tests
+
+- `TestOverlayComponentState_*` — 6 unit tests covering add/replace/leave-alone/single-component/nil-slip/pipeline-level-step cases
+- `TestUpdateAggregateStatusFromComponentStates_AsyncInsertRace` — original regression test for in-memory Aggregates correctness
+- `TestOverlayUpdatesStepsStatus_T1Regression` — explicitly asserts `slip.Steps["builds"].Status == completed` post-overlay (would fail without the recompute block)
+- `TestOverlayNewComponentUnderCanonicalKey_T2Regression` — asserts new components land in canonical bucket, not alias bucket
+
+#### Spec alignment (`STATE_MACHINE_V3.md`)
+
+| Invariant | Impact |
+|---|---|
+| I1 (aggregate from event log) | Preserved — overlay row IS the just-appended event |
+| Append-only events | Preserved |
+| Terminal-success monotonic | Preserved |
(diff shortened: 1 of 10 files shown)


pkuzmenko-mycarrier:
What's broken in PR #59

  The fix doesn't actually fix the bug. Verified:

  Update(slip) → insertAtomicUpdateWithVersions → BuildStepColumnsAndValues(slip.Steps) writes the builds_status Enum8 column from slip.Steps["builds"].Status.

  But overlayComponentState only mutates slip.Aggregates["builds"][i].Status — never touches slip.Steps["builds"].

  So after overlay:
  - slip.Aggregates["builds"] → correct (completed)
  - slip.Steps["builds"].Status → still running (set by hydrateSlip from stale CH read)
  - DB write: builds_status = running (from Steps map)
  - DB write: aggregate JSON column (from Aggregates) → correct

  → Bug survives. The aggregate JSON is right, but the scalar status column (what downstream code actually checks for completion) stays stuck at running.

  Fix needed

  After overlayComponentState, recompute aggregate from slip.Aggregates[aggregateStepName] and apply to slip.Steps[aggregateStepName] via ApplyStatusTransition. Plus T2 (use canonical aggregate key in !updated branch). Plus T3 (extend
  regression test to assert slip.Steps["builds"].Status AND drive updateAggregateStatusFromComponentStates via mock session to catch serialization gap).

(c) pr-review skill

bcarlock-mycarrier (approved):

PR #72, opened 2026-06-19
feat(slippy): I5 race fix — Option 1 (monotonicity gate + flag) (ADO #82468)

pkuzmenko-mycarrier:
## Summary

Closes the I5 materialization-violation race class that wedged slip
`436cc68c-0dd8-4d9e-a899-c831f3106ea2` (production trigger) and 9 other
slips fleet-wide over the 90-day RCA window. Pairs with `slippy-api` PR
(forthcoming) which wires the gate into the HTTP layer and adds a
per-correlationID lock.

ADO: https://dev.azure.com/IntegratedTM/MyCarrier/_workitems/edit/82468

## Design

Two layers, surviving 3 rounds of devils-advocate review:

1. **Option 1 — Terminal monotonicity at INSERT.** Pre-flight argMax
   query against `slip_component_states` event log refuses non-terminal
   transitions when a prior terminal exists, with explicit recovery
   allow-list (`aborted → pending` cascade-reset per `executor.go:376`;
   `{failed/aborted/error/timeout/skipped} → completed` recovery).
2. **Option D primitives (kept from prior R1+R2 work):** `StepStatusOverride`
   exported type, variadic `Update`, `LatestStepStatusFromEvents` method,
   shared sort-key constants.

## What's in this PR

- `slippy/errors.go` — `ErrTerminalAlreadyExists` sentinel (errors.Is-friendly)
- `slippy/clickhouse_store.go`:
  - `isRecoveryAllowed(prior, incoming) bool` encoding 81-cell state-machine matrix
  - `gateBypassSteps` map (defensive `push_parsed` bypass; rationale in code)
  - `enforceTerminalMonotonicity` helper (sub-ms point lookup via canonical sort-key)
  - Gates wired at `UpdateStep` + `UpdateStepWithHistory`
  - `SetComponentImageTag` exempt (out-of-band metadata write, different bug class)
  - `gateEnabled()` short-circuit via `SLIPPY_I5_GATE_ENABLED` env var (default ON, fail-safe)
  - Fail-OPEN on CH errors with structured WARN log
- Tests:
  - Unit suite covering 81-cell matrix slices, cascade-reset, push_parsed bypass, CH-error fail-open
  - Integration suite `TestI5_Option1_InsertComponentStateGate` w/ 20+ subtests under `//go:build integration` tag
  - Cross-second variant probe (the R1+R2 hole that devils-advocate empirically reproduced)
  - Negative control hard-fails via `t.Fatalf` (prior `t.Logf` silently masked test-env regressions)
  - `os.Setenv → t.Setenv` migration (9 sites)
  - `init()` env-set → `TestMain` migration
- Doc cite fix: `executor.go:377` → `:376`

## Rollback contract (§G.1)

`SLIPPY_I5_GATE_ENABLED=false` — set on goLib-consuming services to disable
the gate without code revert. Default ON (empty/unparseable env → ON).
Symmetric with the slippy-api side `SLIPPY_I5_LOCK_ENABLED` flag.

## Test plan

- [x] `make lint PKG=slippy` clean (0 issues)
- [x] `make test PKG=slippy` PASS (coverage 84.1% slippy, 71.8% slippytest)
- [x] `make check-sec PKG=slippy` (govulncheck) — no vulnerabilities
- [x] Integration: `go test -race -tags=integration ./slippy/...` PASS (CH 25.8 testcontainer w/ async_insert=1, wait_for_async_insert=1)
- [x] Cross-repo composition validated via slippy-api worktree consuming via local replace; full integration suite green there too
- [x] 3-agent mediated review (technical-reviewer + security + devils-advocate, 2 rounds w/ DA empirical probes)
- [x] No regression in 391 prior slippy tests

## Reports (artifact trail)

All under `standup-notes/2026/06/`:
- `rca-stuck-slip-436cc68c-*.md` — RCA chain
- `resolve-i5-option1-stage2-plan-v3.md` — design (DA-approved with minor mods)
- `resolve-i5-option1-stage3a-finisher.md` — Stage 3 impl
- `resolve-i5-option1-stage4-validation.md` — gates
- `resolve-i5-option1-stage6a-golib.md` — B3 closure

## Sequencing

This PR must merge first. After merge + tag (`slippy/v1.3.91` per smart-commits MINOR bump), slippy-api PR can swap its `replace` directive to the published tag.

## Code Review Addendum — Operator Comms & Layered Defense Context

**Layer context.** This PR ships layer-1-of-3 of the I5 terminal-monotonicity defense. Layer 1 is the pre-flight argMax gate at INSERT (THIS PR). Layer 2 is the slippy-api R1 overlay event-log guard (slippy-api PR #39). Layer 3 is the slippy-api per-correlationID Dragonfly lock (slippy-api PR #39, default OFF via `SLIPPY_I5_LOCK_ENABLED`). The gate ALONE does NOT close the 436cc68c bug class — it closes the dominant case but leaves a residual cross-second race window addressed by lock + overlay. Reference: `PROJECT_STATE.md` in `.github/` for full architecture diagrams.

**Operator comms.** `SLIPPY_I5_GATE_ENABLED` defaults ON (fail-safe rollback contract). The first deploy of this lib version will surface 409 (`ErrTerminalAlreadyExists`) responses that were previously silent late-write data corruption. Expected: increased 409 rate from concurrent writers; this is correct behavior — refusals indicate the gate caught a regression that would have corrupted `ci.routing_slips`. Rollback: set `SLIPPY_I5_GATE_ENABLED=false` (env-var, hot kill-switch). Operators should watch for a spike in 409 rate per the slippy-api dashboard panel and correlate with Argo workflow retries.

**CI gap acknowledged.** Integration tests (`clickhouse_store_option1_integration_test.go`, `clickhouse_store_i5_async_insert_integration_test.go`) are behind `//go:build integration`. The current CI workflow (`.github/workflows/ci.yml`) does NOT run integration tests. 81-cell matrix verification + async-insert visibility proofs ride on developer discipline. Follow-up: wire an integration-test job (nightly or label-triggered) — to be tracked separately.

### DIFF
--- a/.github/PROJECT_STATE.md
+++ b/.github/PROJECT_STATE.md
@@ -1374,18 +1374,40 @@ Rationale: non-terminal priors have no monotonicity invariant to violate.
 
 ```text
    prior \ incoming | pending                 | held    | running
-   -----------------+-------------------------+---------+---------
+   -----------------+-------------------------+---------+----------------------
    completed        | REFUSE                  | REFUSE  | REFUSE
-   failed           | REFUSE                  | REFUSE  | REFUSE
-   error            | REFUSE                  | REFUSE  | REFUSE
-   aborted          | ALLOW (cascade-reset)   | REFUSE  | REFUSE
-   timeout          | REFUSE                  | REFUSE  | REFUSE
-   skipped          | REFUSE                  | REFUSE  | REFUSE
+   failed           | REFUSE                  | REFUSE  | ALLOW (Argo retry)
+   error            | REFUSE                  | REFUSE  | REFUSE (no prod data)
+   aborted          | ALLOW (cascade-reset)   | REFUSE  | REFUSE (use cascade)
+   timeout          | REFUSE                  | REFUSE  | ALLOW (Argo retry)
+   skipped          | REFUSE                  | REFUSE  | REFUSE (no prod data)
 ```
 
-`aborted → pending` is the ONE intentional terminal → non-terminal transition,
-driven by `executor.go:377` cascade-reset after a primary failure resolves
-(`TestCheckPipelineCompletion_resolved_failure_resets_cascade_aborted_steps_to_pending`).
+Terminal → non-terminal allowed cells:
+
+1. `aborted → pending` — cascade-reset after primary failure resolves
+   (`executor.go:377`,
+   `TestCheckPipelineCompletion_resolved_failure_resets_cascade_aborted_steps_to_pending`).
+2. `failed → running`, `timeout → running` — Argo workflow-step retry. Argo's
+   workflow-step-level `retryStrategy` re-runs the entire step (pre-job → main
+   → post-job). The pre-job container (Slippy CLI prejob) POSTs
+   `/v1/slips/{id}/steps/{name}/start` to slippy-api, which inserts through
+   this gate. Without these allows the gate returns 409 and Argo's
+   `retryStrategy.expression` (matches only `lastRetry.exitCode ∈ {143, 137,
+   -1}`) does NOT catch the 409 → the workflow step abandons → the slip
+   wedges at failed/timeout. Empirically validated against
+   `workflow-dev/workflows/templates/{auto-deployer.yaml, autotriggertests.yaml,
+   mc-env-tag-reset.yaml, offload-purge.yaml}`.
+
+`aborted → running` is INTENTIONALLY REFUSED: cascade-aborted steps must go
+through `aborted → pending` (rule 1) to re-enter the queue; a direct
+`aborted → running` bypasses the recovery contract.
+
+`error → running` and `skipped → running` are REFUSED based on production
+evidence: a 90-day window of `ci.slip_component_states` (54,473 rows, queried
+2026-06) shows ZERO occurrences of any terminal → running transition,
+confirming neither error nor skipped has a real retry-running pattern. If
+such a pattern surfaces, narrow the allow-list at that point.
 
 ##### Terminal prior × terminal incoming (6 × 6 = 36 cells)
 
@@ -1414,8 +1436,18 @@ failed`) is REFUSED at library level for clean semantics; HTTP layer maps to
      4. error     → completed   (recovery)
      5. timeout   → completed   (recovery)
      6. skipped   → completed   (recovery)
+     7. failed    → running     (Argo workflow-step retry)
+     8. timeout   → running     (Argo workflow-step retry)
 
    ALL OTHER terminal-prior transitions: REFUSE → ErrTerminalAlreadyExists
+
+   NOTE on rules 7-8: Argo's workflow-step retryStrategy re-runs pre-job →
+   main → post-job. Pre-job (Slippy CLI prejob) POSTs /start. Without
+   rules 7-8 the gate returns 409, Argo's retryStrategy.expression catches
+   only signal exit codes (143/137/-1), not 409, and the step abandons →
+   slip wedges. aborted/error/skipped → running remain REFUSED
+   (aborted uses cascade-reset; error/skipped have zero production
+   occurrences in a 90-day window).
 ```
 
 ### Rollback flag matrix (`SLIPPY_I5_GATE_ENABLED`)
--- a/.github/PROJECT_STATE.md
+++ b/.github/PROJECT_STATE.md
@@ -1455,11 +1455,19 @@ failed`) is REFUSED at library level for clean semantics; HTTP layer maps to
 ```text
    GATE flag | Behavior                                | Use case
    ----------+-----------------------------------------+-------------------------------
-   unset     | DEFAULT-OFF — gate returns nil silently | Rollback / pre-cutover
-   "false"   | Gate returns nil silently               | Explicit disable
-   "true"    | Gate enforces matrix; returns 409 on    | Production after Stage-3
-             | refused transitions                      | validation
+   unset     | DEFAULT-ON (fail-safe) — gate enforces  | Production default
+             | matrix; returns 409 on refused          |
+             | transitions                             |
+   "false"   | Gate returns nil silently               | Rollback / explicit disable
+   "true"    | Gate enforces matrix; returns 409 on    | Explicit enable (equivalent
+             | refused transitions                     | to unset)
 ```
+
+`gateEnabled()` (clickhouse_store.go) treats `SLIPPY_I5_GATE_ENABLED` as a
+fail-safe boolean: only the explicit string `"false"` disables the gate; any
+other value, including unset, leaves it enforcing. Earlier drafts of this
+doc described `unset` as DEFAULT-OFF — that was wrong. The deployed default
+is ON.
 
 Gate fails OPEN on transport/lookup errors (logged via `slog.WarnContext` with
 `correlation_id`, `step`, `component`, `error`). This prevents a CH outage
--- a/.github/PROJECT_STATE.md
+++ b/.github/PROJECT_STATE.md
@@ -1268,3 +1268,185 @@ Discovered April 29, 2026. Full detail and fix status in `slippy/DISCREPANCIES.m
 - [ ] CLI tooling for slippy operations (manual slip inspection, cleanup)
 - [ ] Additional logger adapters beyond Zap
 - [ ] Implement `normalizeRepository()` for fork handling
+
+---
+
+## I5 race resolution — goLib monotonicity gate (Layer 1)
+
+ADO #82468 — terminal-monotonicity invariant enforced at the `slippy/clickhouse_store.go`
+INSERT boundary. This file documents the *library-layer* contribution; the slippy-api PR
+documents the full three-tier architecture (lock + R1 overlay + this gate).
+
+### Pre-fix bug (slip 436cc68c — production race)
+
+```text
+                  Pod A (ci_runner_1)                 Pod B (ci_runner_2)
+                  -------------------                 -------------------
+   t0             POST /v1/.../complete (unit_tests)
+                    │
+                    │  store.UpdateStep(corrID,
+                    │       "unit_tests", "", COMPLETED)
+                    │    └─ insertComponentState(...)
+                    │         INSERT ... wait_for_async_insert=0
+                    │         (queued; not yet visible)
+   t1                                                  POST /v1/.../start (unit_tests)
+                                                          │  store.UpdateStep(corrID,
+                                                          │       "unit_tests", "",
+                                                          │       RUNNING)
+                                                          │    └─ insertComponentState(...)
+                                                          │         INSERT ... (queued)
+   t2             (Pod A INSERT flushes)
+   t3                                                  (Pod B INSERT flushes; same µs)
+
+   POST-flush slip_component_states (event log):
+     ts=t2  status=completed  (Pod A)
+     ts=t3  status=running    (Pod B)   <-- argMax winner!
+
+   ROUTING TABLE
+     routing_slips.unit_tests_status = running   (terminal regression)
+     routing_slips.status            = in_progress   (slip stuck forever)
+```
+
+**Root cause:** ClickHouse async-insert visibility race + no INSERT-time
+serialization. Two concurrent writers each read an empty/stale event log, both
+INSERT, and `argMax(ts)` deterministically picks the later (non-terminal) write
+as the "current" status — flipping a terminal `completed` step back to
+`running`.
+
+### Layer 1 — `enforceTerminalMonotonicity` decision flow
+
+Gate is invoked from `UpdateStep` (`clickhouse_store.go:612`) and
+`UpdateStepWithHistory` (`:673`) BEFORE `insertComponentState`. `SetComponentImageTag`
+(`:881`) is intentionally exempt (out-of-band metadata).
+
+```text
+                  enforceTerminalMonotonicity(corrID, step, component, incoming)
+                  ─────────────────────────────────────────────────────────────
+                                          │
+                          SLIPPY_I5_GATE_ENABLED ?
+                                          │
+                            ┌─────────────┴─────────────┐
+                           no                          yes
+                            │                            │
+                       return nil               gateBypassSteps[step] ?
+                       (rollback)                        │
+                                            ┌───────────┴───────────┐
+                                          yes (push_parsed)         no
+                                            │                       │
+                                       return nil          latestComponentStateStatus()
+                                       (§B.15 bypass)                │
+                                                         ┌──────────┴──────────┐
+                                                      err / not-found        found
+                                                         │                    │
+                                                  return nil           prior.IsTerminal() ?
+                                                  (first-write or fail-open)  │
+                                                                ┌─────────────┴─────────────┐
+                                                               no                          yes
+                                                                │                            │
+                                                          return nil          isRecoveryAllowed(prior, incoming) ?
+                                                          (non-terminal prior:                 │
+                                                          always allowed)        ┌────────────┴────────────┐
+                                                                                yes                       no
+                                                                                  │                        │
+                                                                            return nil         return ErrTerminalAlreadyExists
+                                                                            (cascade-reset    (caller maps to 409 in
+                                                                            or recovery)      slippy-api mapWriteError)
+```
+
+#### 81-cell allow-list matrix (StepStatus × StepStatus, full 9×9)
+
+Non-terminal: `pending`, `held`, `running`. Terminal: `completed`, `failed`,
+`error`, `aborted`, `timeout`, `skipped`.
+
+##### Non-terminal prior × all incoming (3 × 9 = 27 cells) — ALL ALLOW
+
+```text
+   prior \ incoming | pending | held | running | completed | failed | error | aborted | timeout | skipped
+   -----------------+---------+------+---------+-----------+--------+-------+---------+---------+--------
+   pending          | ALLOW   | ALLOW| ALLOW   | ALLOW     | ALLOW  | ALLOW | ALLOW   | ALLOW   | ALLOW
+   held             | ALLOW   | ALLOW| ALLOW   | ALLOW     | ALLOW  | ALLOW | ALLOW   | ALLOW   | ALLOW
+   running          | ALLOW   | ALLOW| ALLOW   | ALLOW     | ALLOW  | ALLOW | ALLOW   | ALLOW   | ALLOW
+```
+
+Rationale: non-terminal priors have no monotonicity invariant to violate.
+
+##### Terminal prior × non-terminal incoming (6 × 3 = 18 cells)
+
+```text
+   prior \ incoming | pending                 | held    | running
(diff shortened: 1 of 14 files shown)


pkuzmenko-mycarrier:
**Recovery allow-list fix — critical post-merge gap closed (6d1ba85)**

Post-merge review caught a real Argo-retry regression: `isRecoveryAllowed` refused `failed → running` and `timeout → running`. Argo's workflow-step `retryStrategy` re-runs the entire step (pre-job → main → post-job); pre-job (Slippy CLI prejob) POSTs `/v1/slips/{id}/steps/{name}/start` to slippy-api → gate returned `ErrTerminalAlreadyExists` → HTTP 409 → Argo `retryStrategy.expression` matches only `lastRetry.exitCode ∈ {143, 137, -1}` (SIGTERM/SIGKILL/network), NOT 409 → workflow step abandoned → slips wedged at failed/timeout.

Fix extends the allow-list per production data + design intent. Sibling slippy-api PR #39 unaffected (gate lives in goLib).

**Decisions:**
| Transition | Decision | Rationale |
|---|---|---|
| failed → running | ALLOW | Most common Argo workflow-step retry path |
| timeout → running | ALLOW | CI-typical retry-with-extension |
| aborted → running | REFUSE | Must go through `aborted → pending` cascade-reset |
| error → running | REFUSE | Zero 90d production occurrences (`ci.slip_component_states`, 54,473 rows) |
| skipped → running | REFUSE | Zero 90d production occurrences |

**Changes:**
- `slippy/clickhouse_store.go` — `isRecoveryAllowed` new rule 3 (failed/timeout → running) with full Argo retryStrategy rationale in godoc
- `.github/PROJECT_STATE.md` — 81-cell matrix updated (terminal × non-terminal section + summary)
- `slippy/terminal_monotonicity_gate_test.go` — 4 new predicate + 4 new enforcement tests; narrowed existing terminal-incoming-non-terminal refused test
- `slippy/clickhouse_store_option1_integration_test.go` — subtests 7a/7b/7c covering allow + refuse paths; subtest 7 (aborted → running refused) preserved as regression guard

**Gates:** `make lint PKG=slippy` clean, `make test PKG=slippy` 84.2% coverage (isRecoveryAllowed 100%, enforceTerminalMonotonicity 100%); integration `TestI5_Option1_InsertComponentStateGate` PASS (all subtests incl. new 7a/7b/7c).

.github/PROJECT_STATE.md
```diff
@@ -1268,3 +1268,217 @@ Discovered April 29, 2026. Full detail and fix status in `slippy/DISCREPANCIES.m
...
+   occurrences in a 90-day window).
+```
+
+### Rollback flag matrix (`SLIPPY_I5_GATE_ENABLED`)
+
+```text
+   GATE flag | Behavior                                | Use case
+   ----------+-----------------------------------------+-------------------------------
+   unset     | DEFAULT-OFF — gate returns nil silently | Rollback / pre-cutover
+   "false"   | Gate returns nil silently               | Explicit disable
```
bcarlock-mycarrier: **Rollback matrix contradicts the code: `unset` is documented DEFAULT-OFF but the gate defaults ON.**

`gateEnabled()` (`clickhouse_store.go:1067-1077`) returns `true` when `SLIPPY_I5_GATE_ENABLED` is unset; the godoc says "empty or unparseable → fail-safe ON" and `TestGateEnabled_DefaultsOn` asserts it. The code is internally consistent — this matrix row is the bug. It matters because this is the operator-facing cutover/rollback reference: "unset = DEFAULT-OFF" leads an operator to deploy a consuming service believing the gate is inert, when it is live on first deploy and returns `ErrTerminalAlreadyExists` (→ 409) on every refused transition. Fail-safe-ON is correct for an integrity gate; fix the doc.

```suggestion
   unset     | DEFAULT-ON (fail-safe) — gate enforces   | Production default
   "false"   | Gate returns nil silently               | Rollback / explicit disable
```

slippy/clickhouse_store.go
```diff
@@ -792,6 +900,236 @@ func (s *ClickHouseStore) SetComponentImageTag(
...
+	// Plan v3 §G.1 env-flag kill switch (rollback contract).
+	if !gateEnabled() {
+		return nil
+	}
+	// Step-level bypass (defensive, see gateBypassSteps godoc).
+	if gateBypassSteps[stepName] {
+		return nil
+	}
+
+	prior, found, err := s.latestComponentStateStatus(ctx, correlationID, stepName, componentName)
```
bcarlock-mycarrier, line 1176: **Scope/accuracy: this gate is check-then-act (TOCTOU); on its own it narrows but cannot close the I5 race.**

`enforceTerminalMonotonicity` reads prior status here (`latestComponentStateStatus`, line 1107) and the caller then INSERTs outside any transaction or lock (`UpdateStep`/`UpdateStepWithHistory`). Two concurrent writers for the same `(correlation_id, step)` can both read `prior=empty`/non-terminal and both pass — the window the RCA describes. The PR's own integration tests concede this (`clickhouse_store_option1_integration_test.go:416-417, 468-470`: "the gate alone CANNOT serialize same-µs concurrent INSERTs … that is the lock's job"), and the actual per-correlationID lock lives in slippy-api PR #39.

No code change requested — the gate is sound as defense-in-depth. The ask: release notes / downstream consumers should not treat this gate alone as "the I5 fix." It refuses *settled* terminal regressions (prior INSERT already flushed and visible) but does not fix the same-microsecond race that wedged slip 436cc68c; that residual window closes only when #39 ships the lock.

slippy/clickhouse_store.go
```diff
@@ -792,6 +900,236 @@ func (s *ClickHouseStore) SetComponentImageTag(
...
+	// above for full rationale. aborted/error/skipped → running remain REFUSED.
+	if incoming == StepStatusRunning {
+		switch prior {
+		case StepStatusFailed, StepStatusTimeout:
+			return true
+		case StepStatusAborted, StepStatusError, StepStatusSkipped,
+			StepStatusCompleted, StepStatusPending, StepStatusHeld, StepStatusRunning:
+			// aborted → running: must go through aborted → pending cascade-reset.
+			// error / skipped → running: zero production occurrences (90d, 2026-06).
+			// completed → running: completed is absolutely final.
```
bcarlock-mycarrier, line 1109: **`error → running` is refused, but `STATE_MACHINE_V3.md` classifies `error` as a recoverable primary failure equivalent to `failed`.**

Rule 3 allows `failed → running` and `timeout → running` (Argo step retry) but excludes `error → running` here on 90d-empirical grounds. However `STATE_MACHINE_V3.md:15,20` group `{failed, error, timeout}` as one "primary failure" class and the recovery rule (`:395`) explicitly contemplates an `error` step → `running`. If the platform ever emits an `error` step, an Argo retry POSTs `/start` → 409 → `retryStrategy.expression` (matches only exit `{143,137,-1}`) misses it → the step abandons and the slip wedges at `error` — precisely what Rule 3 prevents for `failed`/`timeout`.

`CLAUDE.md` corroborates `error` is effectively never written today, so this is latent. But it is a load-bearing empirical assumption diverging from the documented contract. Either include `error → running` for symmetry, or add a comment cross-referencing `STATE_MACHINE_V3.md:15` marking the divergence as deliberate.

slippy/clickhouse_store.go
```diff
@@ -374,6 +406,55 @@ func (s *ClickHouseStore) LoadLiveByCommit(ctx context.Context, repository, comm
...
+		}
+		return "", false, fmt.Errorf("LatestStepStatusFromEvents: %w", err)
+	}
+	if statusStr == "" {
+		// argMax can yield empty string when the GROUP BY collapses to a single
+		// row whose status column is the zero value. Treat as "no signal".
+		return "", false, nil
+	}
+	return StepStatus(statusStr), true, nil
+}
```
bcarlock-mycarrier, line 431: **`LatestStepStatusFromEvents` duplicates `latestComponentStateStatus` — delegate instead.**

This public method (424-456) and the private `latestComponentStateStatus` (920-950) share the same argMax/GROUP BY query, sort-key constant, nil-row guard, `sql.ErrNoRows` handling, empty-string→no-signal handling, and wrapping. The only difference is the component predicate (`component = ''` literal vs `component = ?` bound). The public method is the private one with `componentName == ""`, so it can delegate — removing the drift hazard the PR was careful to kill for the sort-key constants.

Caveat: delegation changes this method's error prefix from `LatestStepStatusFromEvents:` to `latestComponentStateStatus:` — observable only if a caller string-matches the error (slippy-api's overlay guard fails-open on query error and should not, but confirm).

```suggestion
func (s *ClickHouseStore) LatestStepStatusFromEvents(
	ctx context.Context, correlationID, step string,
) (StepStatus, bool, error) {
	// Pipeline-level rows have component="" — delegate to the component-aware
	// companion to avoid duplicating the argMax point-lookup and its scan/error
	// handling. See latestComponentStateStatus.
	return s.latestComponentStateStatus(ctx, correlationID, step, "")
}
```

slippy/interfaces.go
```diff
@@ -85,6 +93,23 @@ type SlipStore interface {
...
+	// row exists; returns ("", false, nil) when none exist (caller treats this
+	// as "no event yet, do not block"). Returns ("", false, err) on query
+	// failure — callers should treat this as fail-open (log + continue) so
+	// transient ClickHouse outages do not block CI traffic.
+	//
+	// This method exists to support the I5 race fix in slippy-api's
+	// overlayPipelineStep: callers need to know event-log truth for a step
+	// (not the in-memory snapshot from Load) when deciding whether to apply
+	// a non-terminal overlay on top of a terminal status.
+	LatestStepStatusFromEvents(
```
bcarlock-mycarrier, line 109: **`LatestStepStatusFromEvents` is added to the `SlipStore` interface with no in-module consumer.**

This new interface method (109) and the widened variadic `Update(..., overrides ...StepStatusOverride)` (56) have no non-test caller inside the slippy package — the gate uses the *private* `latestComponentStateStatus`, not this interface method. The sole real consumer is slippy-api's `overlayPipelineStep` in a separate module/PR (#39). Every `SlipStore` implementer (both in-repo mocks, plus downstream test doubles) must now implement it regardless of need.

Deliberate coordinated-rollout choice, and the variadic `Update` is backward-compatible (existing zero-override callers compile unchanged), so not a blocker — flagging because the interface growth is driven entirely by an out-of-tree consumer, worth an explicit nod given the project's YAGNI/ISP stance.

slippy/mock_store_test.go
```diff
@@ -97,8 +97,13 @@ type MockStore struct {
 	UpdateComponentCalls  []UpdateComponentCall
 	AppendHistoryCalls    []AppendHistoryCall
 	SetImageTagCalls      []SetImageTagCall
-	UpdateSlipStatusCalls []UpdateSlipStatusCall
-	CloseCalls            int
+	UpdateSlipStatusCalls          []UpdateSlipStatusCall
+	LatestStepStatusFromEventsCalls []LatestStepStatusFromEventsCall
```
bcarlock-mycarrier, line 101: **`mock_store_test.go` is not gofmt-clean after the new mock fields were added.**

`gofmt -l slippy/mock_store_test.go` flags this file — adding `LatestStepStatusFromEventsCalls` (the longest field) widened the struct's alignment column, but the surrounding fields (`CreateCalls`…`CloseCalls`, lines 89–102) were not realigned and `CloseCalls` lost a space. `make fmt` (`golangci-lint fmt`) will rewrite this block, showing up as an uncommitted diff against the project's format gate.

Note this will not fail the CI lint job — `.golangci.yml` sets `run.tests: false` so `golangci-lint run` skips `_test.go`, and there is no format-diff check in `ci.yml`. Local hygiene, not a CI blocker. A GitHub suggestion can't fix it inline (the misaligned run starts at line 89, outside this hunk) — run `gofmt -w slippy/mock_store_test.go` (or `make fmt`) before merge.

slippy/clickhouse_store.go
```diff
@@ -792,6 +900,236 @@ func (s *ClickHouseStore) SetComponentImageTag(
...
+
+// isRecoveryAllowed encodes the explicit allow-list of terminal → X transitions
+// permitted by the gate. Returns true only if the (prior, incoming) pair is
+// on the allow-list.
+//
+// Allow rules (precondition: prior is terminal):
+//
+//  1. prior == aborted AND incoming == pending
+//     → cascade-reset after resolved upstream failure (see executor.go:376,
```
bcarlock-mycarrier: **Stale line citation: `executor.go:376` should be `:377`.**

This godoc cites the cascade-reset at `executor.go:376`, but the actual `UpdateStepWithStatus(... StepStatusPending ...)` call is at `executor.go:377` (line 376 is the `for` loop header) — and `errors.go:69,78` already cite `:377` for the same transition. Wrong and internally inconsistent within this PR.

```suggestion
//     → cascade-reset after resolved upstream failure (see executor.go:377,
```

bcarlock-mycarrier (changes requested):

.github/PROJECT_STATE.md
pkuzmenko-mycarrier: Confirmed — the matrix row was wrong. Code defaults to fail-safe ON when unset (per `TestGateEnabled_DefaultsOn`). Fixed in `bea5930`:
- Updated the matrix row to show `unset OR unparseable → DEFAULT-ON (fail-safe)`.
- Added a 4th implicit case in the polish pass (`2b2f3fe`) — the row label now reads `unset OR unparseable` so the table is self-contained, and the explanatory paragraph below cites `strconv.ParseBool` semantics directly ("any ParseBool-false value disables: `false`, `0`, `False`, `f`, …") rather than just the bare string `"false"`.
- Followup from internal DA review: SMV3 cite at `clickhouse_store.go` line ~1061 also expanded from `15,20` → `15,20,47,395` so the recovery-transition contract is actually findable from the breadcrumb.

The operator-facing risk you flagged (deploy believing the gate is inert when it's live + emitting 409s) is the more important fix here — pre-deploy comms about the expected 409 spike are now in the PR description addendum.

slippy/clickhouse_store.go
pkuzmenko-mycarrier: Agreed — gate is check-then-act and cannot close the residual same-µs race on its own. We treat this as layer 1 of 3 in defense-in-depth:
- **Layer 1 (this PR)** — pre-flight argMax gate refuses *settled* terminal regressions; closes the dominant case where async-insert has flushed.
- **Layer 2** — R1 overlay event-log guard in slippy-api `overlayPipelineStep` (slippy-api PR #39); consults event log when caller-status and event-log-status are both terminal and differ, pins event-log status into the override.
- **Layer 3** — per-correlationID Dragonfly lock in slippy-api (`SLIPPY_I5_LOCK_ENABLED`, default OFF, gated on Slippy CLI PR3 for 409 retry-with-jitter).

Architecture diagrams are in `.github/PROJECT_STATE.md`. PR description addendum makes the layered-defense scope explicit so downstream consumers don't treat the gate alone as the I5 fix. No code change requested here.

slippy/clickhouse_store.go
pkuzmenko-mycarrier: Good catch on the SMV3 divergence. We deliberately diverge from the spec's `{failed, error, timeout}` symmetry and document the reason. Two changes landed:
1. **Code unchanged** — `error → running` still refuses. Rationale: 90d ClickHouse sweep of `ci.slip_component_states` (54,473 rows) returned 0 occurrences of `error` status writes in production. Adding it later is a one-line allow-list extension if observability ever surfaces a legitimate case.
2. **Comment added** (commit `cc64639`, expanded in `2b2f3fe`) — cites `STATE_MACHINE_V3.md:15, 20, 47, 395` so a reader following the breadcrumb lands on the actual recovery-transition contract (line 47 + symmetric branch at 395), not just the glossary. Comment explicitly flags the divergence as deliberate-empirical, not an oversight.

If you have telemetry showing legitimate `error` writes happening today (or expect to), happy to add the allow rule preemptively — your call.

slippy/clickhouse_store.go
pkuzmenko-mycarrier: Accepted. Done in `3d7a22a` — `LatestStepStatusFromEvents` now delegates to `s.latestComponentStateStatus(ctx, correlationID, step, "")` exactly as you suggested. Drops ~30 lines and the sort-key-formula drift hazard you flagged elsewhere.

On the consumer-prefix caveat — verified slippy-api `slip_writer.go` only does `err.Error()` for log/span fields, no `errors.Is`/`strings.Contains` against the old `LatestStepStatusFromEvents:` prefix. Error wrap chain is preserved (`fmt.Errorf("%w", ...)` style throughout). The new prefix is `latestComponentStateStatus:` — same `errors.Is` matching behavior.

Side-effect from internal nit pass (`2b2f3fe`): the delegate's nil-row error format used `%s/%s/%s` which produced a trailing slash when component is empty — switched to `%q` so empty renders as `component=""` rather than dangling.

slippy/interfaces.go
pkuzmenko-mycarrier: Fair flag. Acknowledged: `LatestStepStatusFromEvents` on the `SlipStore` interface has no in-module caller. The only real consumer is slippy-api's `overlayPipelineStep` in PR #39. We accepted this as a coordinated-rollout cost because:
- The variadic `Update(..., overrides ...StepStatusOverride)` is backward-compatible (zero-override callers compile unchanged).
- The new interface method needs to be on `SlipStore` (not a separate narrower interface) because slippy-api consumes the full store — splitting would force composition gymnastics in the consumer.
- Both in-repo mocks (`slippy/mock_store_test.go`, `slippy/slippytest/mock_store.go`) are updated; the YAGNI cost lands only on out-of-tree implementers, of which slippy-api is the sole non-test caller today.

If slippy-api PR #39 lands as expected, the interface method gets a real consumer immediately and the temporary YAGNI footprint resolves. If it stalls, we'd revisit.

slippy/mock_store_test.go
pkuzmenko-mycarrier: Already fixed locally before the review landed — see `cf50c17` (`style(slippy): gofmt mock_store_test.go field alignment`). `gofmt -l slippy/mock_store_test.go` returns empty on current HEAD. Confirmed CI lint masks this because of `tests: false` in `.golangci.yml`, but `make fmt` is now clean on this file.

slippy/clickhouse_store.go
pkuzmenko-mycarrier: Already fixed in `dbcdc33` (`docs(slippy): replace executor.go line pins with function-name refs`) — to address the broader line-rot problem you flagged. The godoc now reads "cascade-reset loop in `Client.checkPipelineCompletion` in `executor.go`" rather than line-pinning. `errors.go:69,78` got the same treatment. If a function name ever becomes ambiguous we can switch back to a line ref, but for now function-name + filename is the more durable form.

smunukutla-mycarrier (approved):

bcarlock-mycarrier (approved):

PR #76, opened 2026-07-08
feat(slippy): I5 fix v2 — freshness-window gate + argMax-derived write-back (ADO #83405, re-closes #82468)

pkuzmenko-mycarrier:
## Why v2

v1 (the reverted 26.28 line, PR #72) refused ALL `completed → running` transitions at the event path. That blocked the pipeline restart idiom — the rerunner's only scopes (`all`, `unit_tests`) restart the cascade by re-firing already-completed kickoff steps — forcing devs into bump-commits (ADO #83405). Deployment was rolled back to 26.26.

Deeper: the independent audit showed the v1 gate guarded the wrong path. The 436cc68c corruption travelled through the **aggregate write-back** (`insertAtomicUpdateWithVersions` copying a stale in-memory snapshot), not the event log — the event log never had a stale event.

## What v2 does (branch cut from `32b2546`, pre-v1 — v1 is NOT in this lineage)

**Component 1 — freshness-window gate (event path).** Refuses a NON-terminal write over a terminal step status only when the terminal event is FRESH (default 5s, env `SLIPPY_I5_FRESHNESS_WINDOW_SECONDS`, clamped to 3600s, kill-switch `SLIPPY_I5_GATE_ENABLED` default ON, fail-open on CH errors). Everything else passes: terminal→terminal always (recovery), aborted→pending always (cascade-reset), any rerun/retry seconds-or-later. Age is computed **server-side in ClickHouse** (`dateDiff` against `now64()`) — immune to pod/CH clock skew. Race artifacts land in milliseconds; humans and Argo retries (10–15s backoffs) never hit the window.

**Component 2 — R2 argMax-derived write-back (aggregate path — the actual fix).** `insertAtomicUpdateWithVersions` no longer trusts the in-memory snapshot: pipeline-step columns are derived from a `GROUP BY step` argMax over `slip_component_states` at write time. Precedence: caller override (the just-written step) > argMax-derived > coalesce('pending'). Aggregate steps: the component rollup is authoritative — the derived value wins only in the forward-progress case (derived terminal over in-memory non-terminal = exactly the 436cc68c staleness signature). `Update()` delegates to a single shared init path (`updateWithOverrides`); `AbandonSlip`/`PromoteSlip` gain the same protection.

## Contract for consumers

- No `SlipStore` interface changes. slippy-api needs only a version bump (no code).
- `ErrTerminalAlreadyExists` (maps to HTTP 409 upstream) is now returned ONLY for sub-5s stale writes — reruns are unaffected.

## Testing

- 30+ gate unit tests (boundaries, custom window, clamp, bypass, fail-open, terminal-incoming short-circuit with zero-query assertion)
- 8+ R2 unit tests (3-tier precedence, aggregate merge matrix incl. the stale-pipeline-event-vs-failed-rollup clobber case)
- Integration (`//go:build integration`): `TestI5_V2_436cc68c_Repro` staging the exact production topology (stale routing_slips vs fresh event log, pure + aggregate subtests), SC-1 adversarial, freshness boundary tests. Compile-verified; require Docker CH to run.
- `make lint` 0 issues; `make test` green, 84% coverage.

## Review trail

Plan + implementation went through 2 devils-advocate rounds + technical + security review internally; all blockers fixed in-branch (see commits `430963e`, `86ae73b` for the two DA-round fixes). Plan: monorepo `standup-notes/2026/07/i5-v2-freshness-gate-plan.md`.

## Rollout

1. Merge → release workflow cuts canonical `slippy/v1.3.9x`
2. slippy-api: one-line go.mod bump (separate PR)
3. ManagementInfra: redeploy slippy-api
4. Watch: 409 rate (expect ~0 outside genuine races), rerun success, no I5 divergence

ADO: #83405 (rerun blocked), #82468 (original I5 race).


smunukutla-mycarrier (approved):

pkuzmenko-mycarrier:
Closing this PR deliberately — not because the work is wrong, but because the investigation it triggered outgrew it.

While validating this fix we ran a 90-day production replay (from routing_slips.state_history, the append-only journal) and per-case forensics on every transition the freshness gate would have refused. Findings that reshaped the design: (1) the artifact class (same-request stale write-backs, including one exact 436cc68c signature at 113ms) clusters below ~465ms, while legitimate fast re-runs start at ~1.7s — so the 5s window in this PR is mis-tuned; (2) two previously unknown TestEngine retry storms surfaced (378 and 900-event journals — separate bug, being filed); (3) most importantly, the root cause is architectural: every write-back INSERTs a full-row snapshot computed from a private Load, versioned by land time — the slowest writer's past shadows everyone's present. The gate polices the event path; the disease is in the write-back path.

The consolidated fix (this branch's R2 derive-at-land-time work, PLUS dirty-check write suppression, PLUS a re-tuned ~750ms gate) is specified in the monorepo at `standup-notes/2026/07/slip-state-ch-fix-spec-and-plan.md`, with a 24-item behavioral-contract checklist mined from this repo's change history and a full blast-radius inventory (including the ci_metrics.mv_slip_metrics MV that parses step_details JSON hourly). A fresh implementation will start from this branch's tip (`686146c` — all 13 commits here carry over; they're preserved by the `slippy/v1.4.0-feature-83405-freshness-gate.*` tags after branch deletion). ADO #83405 / #82468.

PR #43, opened 2026-02-24
chore(deps): update goLibMyCarrier to v1.3.63

pkuzmenko-mycarrier:
## Dependency Update

| Field | Value |
|-------|-------|
| Library | `github.com/MyCarrier-DevOps/goLibMyCarrier` |
| Previous version | `../clickhouse` |
| New version | `v1.3.63` |
| Repository | `MyCarrier-DevOps/goLibMyCarrier` |

### Changes

See [releases](https://github.com/MyCarrier-DevOps/goLibMyCarrier/releases) for changelog.

### Checklist

- [ ] `go.mod` updated
- [ ] `go.sum` regenerated via `go mod tidy`
- [ ] Tests pass

---
_Created by go-lib-renovate.sh (local mirror of Copilot Go-Lib Renovate agent)_

PR #77, opened 2026-07-09
feat(slippy): I5 v2 slip-state race fix — R2 + D3 suppression + 750ms freshness gate (ADO #83405)

pkuzmenko-mycarrier:
## Summary

Slip-state ClickHouse race fix, v2 (ADO #83405 / #82468). Implements the settled design in `mycarrier/standup-notes/2026/07/slip-state-ch-fix-spec-and-plan.md` §2:

- **D2 (R2)** — derive-at-land-time: `updateWithOverrides` derives step statuses via a fresh `argMax` query over `slip_component_states` at write time, with 3-tier precedence (caller override > argMax-derived > in-memory) and monotonic-forward merge for aggregate steps. *(Inherited from the reviewed chain `f7ee07d..686146c`, beta tags `slippy/v1.4.0-feature-83405-freshness-gate.*` — 2 design-agent rounds + tech + security review.)*
- **D3 (new)** — dirty-check write suppression: skips the `routing_slips` INSERT (and version bump) when the R2-resolved row is byte-identical to the **persisted** row fingerprinted at Load. Eliminates no-change echo rows — the primary fuel for version-order collisions. (Storm write-volume dampening comes from the D4 gate, not D3 — see round-1 SF-2 below; synthetic storm test: 200 events → 2–4 rows, 59–60 gate refusals.)
- **D4 (changed)** — freshness gate window: `SLIPPY_I5_FRESHNESS_WINDOW_SECONDS` (default 5 s) → `SLIPPY_I5_FRESHNESS_WINDOW_MS` (default **750 ms**; artifact ceiling 465 ms vs legitimate-restart floor 1700 ms from the 90-day replay). Values non-integer / ≤0 / >3,600,000 now **fail at startup** (`NewClickHouseStoreFromConfig`). Kill-switch `SLIPPY_I5_GATE_ENABLED` unchanged.

**No schema changes**: `state_history` structure, `routing_slips` column names, `builds` JSON array shape, and `step_details` JSON shape are all untouched (D1, BC-21/22/25/26 — Grafana panels and `ci_metrics.mv_slip_metrics` contracts preserved).

## Commits on top of reviewed chain tip `686146c`

| Commit | What |
|---|---|
| `9c5afb5` | 750 ms window + MS env var + startup validation |
| `748c779` | D3 dirty-check suppression + BC-13 godoc + 9 unit tests |
| `683cbbd` | Suppression-collapse + synthetic storm-replay integration tests |
| `f4a947a` | Fix: fingerprint baseline captured **pre-hydration** (persisted row, not derived view) — bug caught by the new collapse test + pre-existing `TestE2E_ConcurrentTerminalStepEvents` invariant checker |

## Gates

- `make lint PKG=slippy` — 0 issues
- `make test PKG=slippy` — 470 passed / 0 failed (84.3% coverage)
- `go vet -tags integration ./...` — clean
- Integration suite (`-tags integration -race`, real ClickHouse via testcontainers): all `TestI5V2_*` + `TestE2E_ConcurrentTerminalStepEvents_*` + new suppression/storm tests **PASS**
- `GOTOOLCHAIN=local make check-sec PKG=slippy` (Go 1.26.5 = CI-resolved) — 0 affecting vulnerabilities

## Notes for reviewers (authoring session)

1. **BC-16 / D2 tier-3 divergence (adjudication needed):** spec D2 says "coalesce to `'pending'` if both [override and argMax] absent"; the reviewed chain's `resolveEffectiveStepStatuses` tier 3 falls back to the **in-memory** value instead. Inherited from the reviewed chain, not changed here (changing it would relitigate reviewed behavior). In practice they coincide for freshly-loaded slips (persisted default is `pending`).
2. **§6.1 "0 ms → allow" vs BC-11:** implemented as REFUSE at age 0 per BC-11 (duplicate-write artifact class); documented in the gate code + boundary test.
3. **Startup validation lives only in `NewClickHouseStoreFromConfig`** (slippy-api's actual constructor, main.go:214). `FromSession`/`FromConn` don't return errors today (~130 call sites); not changed.
4. **9 pre-existing `UpdateAggregateStatus*` unit tests re-staged** with real timestamps: their zero-timestamp fixtures made the staged writes genuine no-ops that D3 now (correctly) suppresses.
5. **Pre-existing integration failures, out of scope, untouched:** `TestVersionedCollapsingMergeTree_DataIntegrity` (Load never selects `ancestry` column) and `TestHydrateSlip_EmptyAggregates` (zero `CreatedAt` → 1970 partition). Both reproduce independent of this branch. `TestI5V2_Gate_WithinWindow_Refuses` showed one cold-start flake in a full-suite run early on; passed 6/6 standalone and in later runs.
6. **BC-10 fixture:** boundary tests cover 1700 ms → allow; full `would_refuse_list.jsonl` replay left to the authoring session which owns the dataset.
7. **BC-12:** storm test is a synthetic stand-in for the `fe77f613`/`3af149aa` journals; real journal replay fixture tracked separately per spec.
8. **E2E (§6.3, incl. ADO #83405 rerun acceptance)** pending deploy to argo-events-tests via the slippy-api PR (MyCarrier-DevOps/slippy-api companion PR).

Per spec §8: do not merge without pr-review + robustness-review sign-off from the authoring session.

---

## Review round 1 — addressed (`1716065`)

Findings: `standup-notes/2026/07/i5-v2-review-round1-findings.md`. Verdict was READY after 1 must-fix + 4 should-fix; all applied:

- **MF-1** stale "currently FAILS" doc blocks rewritten (bug was fixed in `f4a947a`, tests are now regression guards)
- **SF-1** BC-13 heal-claim scoped correctly (self-heal only on writes reaching `updateWithOverrides`; N-way all-suppress wedge clause added — §7.2 Query 2 detection)
- **SF-2** storm honesty: volume dampening is D4 gate refusals, not D3 (test comment + spec §7.3 amended)
- **SF-3** `FindByCommits`/`FindAllByCommits` godoc: no fingerprint → D3 inactive, fail-open
- **SF-4** real bug fixed: `freshnessWindow()` now caps ms before the Duration multiply — huge values (1e13) overflowed int64 to a negative window, silently killing the gate on paths bypassing startup validation. +5 tests.

Gates re-run post-fix: lint 0, tests 0 failed, `GOTOOLCHAIN=local` check-sec 0 affecting.

## Tracked follow-ups

- ~~**FU-1** — storm-test assertion strengthening~~ — **CLOSED in round 2** (`27301ab`): storm test now derives the expected aggregate from per-component `argMax` + the production `computeAggregateStatus` rollup and asserts the settled column equals it.
- **FU-2** — D3 suppression observability: promote Debug-only suppression log to Info-with-counter or a metric so operators can distinguish suppression-working from writes-broken (ties to §7.2 row-volume signal). *(Round 2 partially mitigates: the audit-relevant subset — suppression dropping an appended `state_history` entry — now Warns with `state_history_lost: true`.)*
- **FU-3** — F1 adjudication record: N-simultaneous-completer all-suppress wedge is real-but-accepted (own-event visibility via overlayComponentState; prereqs read events; Gap-B reconciler is the systemic net).
- **FU-4** *(new, round 2)* — server-stamp `slip_component_states.timestamp` (`DEFAULT now64(6)`, drop the client timestamp arg in `insertComponentState`) so both gate operands use the CH clock and the freshness gate becomes genuinely skew-immune. Deferred: changes write semantics for every producer of the column; comments now state the residual pod↔CH-server NTP offset precisely instead.

## Test-Environment Validation (argo-events-test, slippy-api-test → ci_test, final build)

The final build of this change was deployed to the test environment (ManagementInfra `slippy-api-test.yaml` image bump) and validated against the exact production scenarios that motivated the fix — including the two the v1 attempt failed on.

**1. Retrigger against the current head commit** ([run 29093403856](https://github.com/MyCarrier-Engineering/MC.Example/actions/runs/29093403856)) — the ADO #83405 acceptance case. The slip's kickoff steps (builds, unit_tests, secret_scan) had been `completed` for ~45 hours. Under v1 this retrigger died with HTTP 409 at the first `/start`. Under this build: both steps re-fired cleanly (`builds` running 12:42:24 → completed 12:43:59; `unit_tests` running 12:42:25 and progressing), **zero 409s anywhere in the chain**, same correlation ID correctly reused, pipeline progressed normally. The freshness gate allowed the transitions because the prior terminal events were ~45h old — far beyond the 750 ms race-artifact window, exactly as the 90-day production replay predicted (legitimate re-runs never come faster than ~1.7 s; race artifacts never slower than ~465 ms).

**2. Retrigger against an outdated commit** ([run 28810817745](https://github.com/MyCarrier-Engineering/MC.Example/actions/runs/28810817745), re-run of an old workflow) — the adversarial ancestry case. The old commit has no slip of its own; slippy-find resolved via ancestry to the nearest ancestor slip, which is in `completed` (fully terminal) state. Outcome: the rerun was accepted upstream but **the completed slip remained immutable** — no state mutation, no step re-fire, no wrong-binding into any newer slip, no wedge, no opaque 409. The slip-level terminal guard held, which is the correct behavior: pre-v1 code would have re-run into the completed slip and overwritten its terminal states.

**Data-integrity signals during both tests** (ci_test): zero divergence between `routing_slips` step columns and the `slip_component_states` event-log argMax (the R2 write-back deriving correctly); `state_history` transition-complete; no stale-running wedges; slippy-api-test started clean (async-insert startup assertion and freshness-window validation both passed).

**One adjacent finding** (pre-existing, not introduced or affected by this PR): when the rerunner's exact-commit slip lookup finds nothing (the outdated-commit case), pushhookparser discards the correlation ID carried in the retrigger payload and the build workflow's `slip-pre` receives an empty parameter — the prejob's resolve-by-ref ancestry fallback then binds correctly, so behavior is safe, but the double-resolution is unnecessary. Tracked separately (pushhookparser `rerunner.go` — adopt payload correlation_id as fallback when lookup is empty).

---

## Review round 2 — addressed (`27301ab`)

Four inline findings from the human review (all threads replied in-line):

- **state_history audit gap** — D3 suppression that would drop a `StateHistory` entry appended since Load now emits the existing `state_history_lost: true` Warn contract (`reason: "d3_suppression"`) instead of a Debug line; baseline `len(StateHistory)` captured pre-hydration alongside the fingerprint, refreshed after each INSERT. +2 unit tests.
- **Clock-skew claim precision** — gate comments no longer claim general pod/CH skew immunity: immune to pod↔pod skew, but the writer-pod↔CH-server NTP offset is present in every evaluation (~tens of ms healthy vs 750 ms window). SQL unchanged; server-stamping tracked as FU-4.
- **Semver break (removed `StepStatusOverride`, narrowed `Update`, removed `LatestStepStatusFromEvents`)** — documented below; org-wide default-branch code search found zero external references (only goLibMyCarrier + slippy-api, which removes its uses in the companion PR #43). Documented-break path chosen over a `/v2` module split.
- **Deterministic convergence-under-race assertion** — added to `TestI5V2_StormReplay_Synthetic` (closes FU-1, see above). Passed 3/3 integration runs under `-race`.

Gates re-run: lint 0; unit 477 passed / 0 failed; integration storm+collapse tests pass; `GOTOOLCHAIN=local` check-sec 0 affecting. Beta tag `slippy/v1.4.0-feature-83405-freshness-gate.16` → `27301ab`; slippy-api pin bumped in PR #43.

## ⚠️ Breaking changes (shipped in `slippy/v1.3.95`)

*(Post-merge note: the canonical tag landed as `slippy/v1.3.95` — the semver tooling computed it from the merged commit types; the spec's `v1.4.0` was an assumption. Strictly these removals are semver-breaking even at a patch tag, but with zero external consumers the practical impact is unchanged. slippy-api PR #43 pins `v1.3.95` as of `2c632b3`.)*

Under semver these removals are breaking; org-wide code search (default branches) confirms no external consumers, so this ships as a documented break rather than a `/v2` module split:

1. **Removed exported type `StepStatusOverride`.**
2. **Narrowed `SlipStore.Update(ctx, slip, overrides ...StepStatusOverride)` → `Update(ctx, slip)`.** Zero-override callers are source-compatible; callers passing overrides must drop them — R2 derives step statuses at land-time from `slip_component_states`, which supersedes caller-supplied overrides.
3. **Removed `SlipStore.LatestStepStatusFromEvents`.** Its read-your-writes role is now internal to the store (`enforceTerminalFreshnessGate` pre-INSERT SELECT + argMax derives).

External `SlipStore` implementers (mocks) must update signatures accordingly. Consumers found importing the module (pr-merge-sync, MC.TestEngine, pushhookparser, imagetagupdater, offload-generator, gitops-rollback, Slippy, slippy-find) reference none of the removed symbols.


slippy/clickhouse_store.go
```diff
@@ -942,261 +1021,276 @@ func (s *ClickHouseStore) SetComponentImageTag(
...
+				"correlation_id": slip.CorrelationID,
+				"error":          fpErr.Error(),
 			})
-		return nil //nolint:nilerr // fail-open is the documented §J error policy
+		} else if currentFP == slip.loadedWriteFingerprint {
+			s.logger.Debug(ctx, "write suppressed: no change since load", map[string]interface{}{
+				"correlation_id": slip.CorrelationID,
+				"suppressed":     true,
+			})
+			return nil
```
smunukutla-mycarrier: **D3 suppression drops a `state_history` journal entry with only a Debug log — no `state_history_lost` signal.**

When `updateAggregateStatusFromComponentStatesWithHistory` appends a caller audit entry to `slip.StateHistory` (`:2860`) and then `updateWithOverrides` suppresses the write (`:1265`), the appended entry is never persisted — `writeFingerprint` excludes `state_history` (`:2087-2093`), so an idempotent, column-no-op transition (e.g. a duplicate terminal→terminal re-report allowed by gate Rule 3 at `:2477`) is suppressed.

This is **not data loss** of the transition itself — `insertComponentState` durably records status/message/timestamp to `slip_component_states` (`:2551`/`:2557`); only the `Actor` field and the denormalized journal entry are dropped, and only for no-net-change transitions. Thinner journals are documented as intended.

The actionable gap is **observability**: every other best-effort write-back logs `state_history_lost: true` at Warn (`:785-795`, `:816-826`) — the monitoring contract the code comments define — but the suppression path logs only Debug (`:1261`). Operators watching for audit gaps silently miss these drops.

_Suggestion_: when suppression fires on a slip whose `StateHistory` was appended since Load, emit the existing `state_history_lost: true` Warn instead of a Debug line. No change to suppression behavior required.

slippy/clickhouse_store.go
```diff
@@ -2254,6 +2398,132 @@ func (s *ClickHouseStore) computeAggregateStatus(componentData []ComponentStepDa
...
+) (status StepStatus, eventTimestamp time.Time, age time.Duration, found bool, err error) {
+	query := fmt.Sprintf(`
+		SELECT
+			argMax(status, %s)                                                   AS latest_status,
+			argMax(timestamp, %s)                                                AS latest_ts,
+			dateDiff('millisecond', argMax(timestamp, %s), now64(6))            AS age_ms
+		FROM %s.%s
+		WHERE correlation_id = ? AND step = ? AND component = ?
+	`, componentEventSortKeyNoImageTag, componentEventSortKeyNoImageTag, componentEventSortKeyNoImageTag,
+		s.database, TableSlipComponentStates)
```
smunukutla-mycarrier, line 2461: **Gate age mixes a client-stamped timestamp with server `now64()` — the "clock-skew-immune" claim is overstated.**

`latestComponentStateRow` computes `age = dateDiff('millisecond', argMax(timestamp), now64(6))`. The `timestamp` column is pod/client-stamped (`insertComponentState :2541`, `time.Now()`) with no server `DEFAULT` (`dynamic_migrations.go:334`), while `now64(6)` is server-evaluated. So `age = server_now − pod_write_time`, i.e. `real_elapsed + (server_offset − pod_offset)`.

The doc's claim of immunity to "pod/CH clock skew" (`:2408-2410`, `:2481`, `:2502-2503`) is overstated: the residual pod↔CH-server NTP offset is present in **every** evaluation — including same-pod writer/evaluator, since the two `dateDiff` operands always come from two different clocks. It's merely small (<~50ms) under healthy NTP against the 750ms threshold; under degraded NTP the effective window shifts and the gate can under-dampen (wrongly allow a race artifact) or over-dampen (wrongly refuse a legitimate write).

_Suggestion_: server-stamp `slip_component_states.timestamp` (column `DEFAULT now64(6)`; drop the client timestamp arg in `insertComponentState`) so both operands use the CH clock and the gate becomes genuinely skew-immune — or soften the comment to state the residual pod↔server offset precisely so operators don't over-trust the 750ms boundary.

slippy/interfaces.go
```diff
@@ -45,15 +45,7 @@ type SlipStore interface {
...
-	//
-	// Optional StepStatusOverride values pin specific <step>_status columns (and their
-	// step_details.<step>.status JSON counterpart) to caller-supplied truth, bypassing
-	// slip.Steps[name].Status for those steps. Zero overrides preserves the historical
-	// behavior — every step column is sourced from slip.Steps[name].Status — and is the
-	// expected mode for callers like AbandonSlip / PromoteSlip that do not transition
-	// individual step states. The override hook exists to defeat the stale-Load race in
-	// slippy-api's hydrateAndPersist (see goLibMyCarrier I5 fix / ADO #82468).
-	Update(ctx context.Context, slip *Slip, overrides ...StepStatusOverride) error
+	Update(ctx context.Context, slip *Slip) error
```
smunukutla-mycarrier, line 48: **Removing exported `StepStatusOverride`, the `Update` variadic, and `LatestStepStatusFromEvents` is a semver-breaking change.**

This PR deletes the exported type `StepStatusOverride`, narrows `SlipStore.Update` from `Update(ctx, slip, overrides ...StepStatusOverride)` to `Update(ctx, slip)` (this line), and removes `LatestStepStatusFromEvents`. `slippy` is a published module tagged `v1.x`, so these are breaking changes under semver. In-repo it's consistent — no dangling references, builds/vets clean — so nothing in this repo breaks.

External impact is the class of consumers that (a) reference `StepStatusOverride`, (b) call `Update` with one or more overrides, or (c) implement the `SlipStore` interface. A plain `Update(ctx, slip)` caller (zero overrides) is unaffected.

Note the remediation is non-trivial in Go: a major bump requires the module path to become `.../slippy/v2` plus consumer import-path updates — not a plain tag bump. Either do a coordinated v2 release with downstream consumers, or, if the removed symbols are confirmed unused downstream, document the break explicitly in the release notes. No code change required in this PR.

slippy/clickhouse_store_i5_v2_integration_test.go
```diff
@@ -0,0 +1,831 @@
...
+	scsRow.Close()
+
+	t.Logf("storm-replay: routing_slips.builds_completed_status=%s (argMax over slip_component_states=%s across all components; last-writer-wins, not necessarily equal to a single component's status)",
+		rsStatus, scsStatus)
+
+	// The routing_slips column must be a status that R2 could plausibly derive from
+	// the component event stream: it must be non-empty and a recognized StepStatus.
+	if rsStatus == "" {
+		t.Errorf("storm-replay: routing_slips.builds_completed_status is empty after storm settled")
+	}
```
smunukutla-mycarrier, line 817: **No deterministic multi-component convergence-under-race assertion exists (author-flagged FU-1).**

The deterministic convergence guard — `TestI5V2_Suppression_CollapseIdenticalWritebacks` (`:580-581`) — is single-component (`"api"`, `:517`/`:547`), so it can't exercise convergence when multiple components race. The multi-component, concurrent test — `TestI5V2_StormReplay_Synthetic` (~40 components, `:600`) — asserts only that the settled status is non-empty (`:805-807`) plus row-count dampening (`:827-829`); it never asserts the settled `routing_slips.<step>_status` equals the value derivable from the event log. So a stale-column wedge at `running` would still pass this test. The author self-flags this as FU-1.

_Suggestion_: after the storm quiesces, assert via direct SQL (like the collapse test) that `routing_slips.<step>_status` equals `argMax(status)` over `slip_component_states` for that step — the convergence-under-race check neither existing test performs today.

(The earlier "would pass on garbage" concern was withdrawn: the `Enum8('pending'=1,…,'skipped'=9)` column constraint at `dynamic_migrations.go:557-562` guarantees any scanned value is a valid status, so the comment/code mismatch is cosmetic.)

smunukutla-mycarrier (changes requested):

slippy/clickhouse_store.go
pkuzmenko-mycarrier: Fixed in 27301ab. The suppression branch now compares `len(slip.StateHistory)` against a baseline captured at the same pre-hydration point as `loadedWriteFingerprint` (and refreshed together with it after every successful INSERT, so a later suppressed write is not falsely flagged for history the previous write already persisted). When suppression fires and the history grew since Load, it emits the existing monitoring contract — `state_history_lost: true` at Warn, with `reason: "d3_suppression"` and the correlation_id — instead of the Debug line; idempotent suppressions with no history growth keep the routine Debug. Suppression behavior itself is unchanged. Two new unit tests pin both paths (Warn fires exactly once with the right fields; no `state_history_lost` on no-growth suppression).

slippy/clickhouse_store.go
pkuzmenko-mycarrier: Fixed in 27301ab (comments only, SQL untouched). All three passages now state the boundary precisely: the server-side `dateDiff(..., now64(6))` makes the gate immune to pod↔pod skew — the two operands never come from two different pods — but every evaluation carries the writer-pod↔CH-server NTP offset, so it is not immune to pod/CH skew in general. Under healthy NTP that residual is ~tens of ms against the 750 ms window; under degraded NTP the effective window shifts and the gate can under- or over-dampen, exactly as you describe.

On the server-stamp alternative: not taken in this PR. A `DEFAULT now64(6)` on `slip_component_states.timestamp` plus dropping the client argument changes write semantics for every producer of that column and interacts with the argMax sort key — too large a design change to fold into a review round on a settled D4. Recorded as a tracked follow-up (FU-4) in the PR body next to FU-2/FU-3.

slippy/interfaces.go
pkuzmenko-mycarrier: No code change, per your note — consumer sweep done to pick between your two options. Org-wide GitHub code search (default branches): zero references to `StepStatusOverride` or `LatestStepStatusFromEvents` outside goLibMyCarrier and slippy-api, and slippy-api removes its own uses in the paired PR #43. Twelve other repos import `goLibMyCarrier/slippy` (pr-merge-sync, MC.TestEngine, pushhookparser, imagetagupdater, offload-generator, gitops-rollback, Slippy, slippy-find, plus docs-only hits) — none reference the removed symbols. Variadic `Update(ctx, slip, overrides...)` callers are covered by the same sweep: constructing an override argument requires naming the `StepStatusOverride` type, and there are zero type hits org-wide. Caveat: code search indexes default branches only, so an unmerged feature branch elsewhere could still reference them — the loud build failure on their next `go get` bump is the backstop.

Given confirmed-unused downstream, taking the documented-break path rather than a `/v2` module split: a "Breaking changes" section now sits in the PR body and will carry into the `slippy/v1.4.0` release notes, listing all three removals and the migration (drop overrides from `Update` calls; R2 derives step statuses at land-time instead).

slippy/clickhouse_store_i5_v2_integration_test.go
pkuzmenko-mycarrier: Fixed in 27301ab — the storm test now performs the deterministic convergence check, implemented the way FU-1's adjudication demanded rather than the naive form: it pulls each component's own `argMax(status, timestamp)` from `slip_component_states` (GROUP BY component, pipeline-level sentinel `''` excluded), feeds those per-component settled statuses through the production rollup itself — `computeAggregateStatus` (clickhouse_store.go:2396, the exact function `updateAggregateStatusFromComponentStates` uses; the integration test lives in package `slippy` so it calls it directly) — and asserts the settled `routing_slips.builds_completed_status` equals that derived value. Deliberately NOT `routing_slips == global argMax` — aggregate rollup and highest-sort-key single event are different quantities, which is why the round-1 devils-advocate rejected the equality form.

Passed 3/3 consecutive integration runs under `-race` (settled `failed` == derived `failed`, 40 components, 200 events). A stale-column wedge at `running` now fails the test. FU-1 is closed by this; PR body updated accordingly.

smunukutla-mycarrier (approved):

PR #78, opened 2026-07-14
fix(slippy): CLONE_DERIVED — derive step columns from event log in atomic clone writers (mycarrier-5dv5)

pkuzmenko-mycarrier:
---
## The bug, in plain English

Slippy stores slip state as an append-only ledger of full-row snapshots: every change inserts a complete new copy of the row with a higher version, and readers take the highest version. Two writers — the status sealer (`insertAtomicStatusUpdate`) and the history appender (`insertAtomicHistoryUpdate`) — build that new copy like a **photocopier**: clone the current top row, change the columns you came to change, stamp a fresh version.

The photocopy is only as good as what the copier can see. ClickHouse inserts from other connections become visible with a small lag, so a sibling handler's just-written step column could be missing from the copy — and because the copy carries the newest version stamp, the **stale value wins the ledger**. Version order guarantees write order, not content freshness.

Normally the next write heals it. The wedge case is when there is no next write: in prod (corr `74cd6676`, 2026-07-13), `unit_tests` failed and sealed the slip terminal while its photocopy hadn't yet seen `secret_scan = completed` (written 0.26 s earlier). Terminal slip → nothing ever writes again → `secret_scan` reads `running` forever. PR #77 fixed the third writer (aggregate write-back) with derive-at-land-time; these two photocopiers were the deliberately-deferred gap (handoff deviation #7), plus one override-free call site (`executor.go` recovery path) that was a plain oversight.

## The fix

Stop trusting the photocopy for step columns — ask the event log instead, inside the same SQL statement. Each pure pipeline step column in the new row is filled by precedence:

1. **Caller's explicit override** (unchanged, highest);
2. **Latest event** from `slip_component_states` — `argMax` by the shared sort key, evaluated server-side at write time;
3. **Photocopied value** — only when the step has no events (documented BC-16 deviation: verbatim, not `pending`).

The INSERT...SELECT stays a single atomic statement — no Load→compute→INSERT round trip, so the `state_history` preservation this path was built for (PR #56/#58) is untouched. Cancel branch and aggregate columns stay byte-verbatim (aggregates are the R2 path's job). The event log is written before the slip rewrite in every handler, and sibling events are typically seconds old — 11 s in the repro — so the derive sees the truth even when the row snapshot lies.

## How we made sure no path is missed

- **Writer census by construction**: every `INSERT INTO routing_slips` writer classified FRESH_ROW / R2_DERIVED / CLONE_DERIVED; a canary test fails CI if a future writer appears unclassified.
- **Regression tests** (testcontainers, live CH 25.8): the exact 74cd6676 interleaving through both writers, override precedence, no-events fallback, cancel-branch integrity, argMax parity with the client-side R2 derive (same shared constant), terminal-seal stability. 8/8 green.
- **Caller walk**: all entry points (`UpdateSlipStatus`, `checkPipelineCompletion` ×3 incl. the override-free recovery site, public `AppendHistory`, `UpdateStepWithHistory`, `handlePushRetry`, aggregate no-rollup branch) verified correct or strictly improved.

## Reviews run before this PR

- **pr-review** (technical + security + devils-advocate deliberation): no blocking findings; applied — backslash-aware escaping + identifier allowlist guard, alignment guard via `StepStatusColumn`, redundant struct field removed, canary wording, doc corrections.
- **robustness-review** (six failure-mode groups + DA): no criticals; applied — `handlePushRetry` self-race closed via `UpdateStepWithHistory`, Warn logging when defensive guards fire, `STATE_MACHINE_V3.md` Rule 9 / Violation 5 amended with the CLONE_DERIVED mechanism and writer taxonomy.
- **robustness-review, second pass on the final branch** (fresh reviewer + DA): one important finding — routing `handlePushRetry` through `UpdateStepWithHistory` silently adopted its best-effort write-back semantics, superseding the a45e63c hard-fail history-append contract, and a divergent test mock masked it. Accepted tradeoff (DA-verified: the event insert still hard-fails, step status self-heals, only a `state_history` audit entry is at risk in a narrow window) and shipped as `775881c`: mock aligned to real store semantics, self-contradictory test rewritten, tradeoff documented in `push.go` and `STATE_MACHINE_V3.md`.
- **Gates**: `make lint` 0 · unit 491/491 · integration regression 8/8 · `check-sec` 0 code findings (7 pre-existing dep advisories in uncalled paths).

## Accepted residuals (documented, not gated)

- P scalar subqueries per write (P = pure steps, correlation-bounded); single-scan LEFT JOIN refactor is a known follow-up.
- If BOTH tables lag simultaneously at a terminal seal, tier-3 falls back verbatim — strict subset of the pre-fix failure surface; Gap-B reconciler territory.
- Fix is inert in prod until slippy-api bumps this module.

Refs: bd `mycarrier-5dv5` · forensic ledger `standup-notes/2026/07/pipeline/i5v2-watch-20260713/ledger.md` · spec `standup-notes/2026/07/slip-state-ch-fix-spec-and-plan.md` §2/§4

## Validation results (2026-07-15, test environment)

The fix was deployed to `slippy-api-test` (image `26.29.70db41e` embedding this branch @ `775881c` via beta tag `.2`) and validated end-to-end against live infrastructure (ClickHouse `ci_test`). Full evidence: `standup-notes/2026/07/slippy-5dv5-validation/verdict.md` (mycarrier repo) + bd `mycarrier-5dv5`.

**What we ran**
- Two full concurrent storms (morning + reproducibility re-run): 12+ dispatches across integration-test (dev + offload/feature13), retrigger-ci, autotrigger-automation-tests, non-prod-deploy — real MC.Example pipelines, real webhook→eventsource→pushhookparser→slippy-api flow.
- Failure injection: Argo workflow `shutdown: Stop` mid-unit_tests → genuine exit-handler failure → slip sealed `failed` → retrigger recovery. (Pod force-deletes proved non-viable — Argo resurrects them without consuming retries; documented.)
- ITU/ancestry probe: retrigger with `target_sha` on auto-bump commits.

**Every fix path exercised live, with the proof signal for each**
| Path | Signal collected |
|---|---|
| Terminal-sibling race (74cd6676 class) | Double unit_tests handler race (completions 82 s / 71 s apart, two storms) — column matched event-log truth both times |
| Same-SHA retry reuse (`handlePushRetry` class) | "Found existing slip for commit, handling retry" — no duplicate slips minted (×2) |
| Supersede/abandon (BC-20) | Chains at 3 abandonment depths; every abandoned slip ended internally consistent |
| Late-completion heal-forward | ×3: step completions landing on already-abandoned slips updated columns WITHOUT reviving terminal status |
| Failure seal (call site 1, overrides) | `primary_failures=[unit_tests]` → slip `failed`; oracle clean at the wedge-prone seal moment |
| Failed→InProgress recovery (executor.go:401, zero-override full-derive) | Fired TWICE ("No primary failures remain, reconciling slip status to in_progress") — second cycle via a genuine test flake; oracle clean after both |
| BC-17/19 rerun semantics | dev_tests/unit_tests completed→running→completed cycles tracked truthfully |

**Aggregate numbers**
- 11+ slips judged (incl. 8+ terminal — the wedge-prone class), 59+ column-vs-event-truth checks: **0 mismatches**
- Version-history walks (all row signs, hundreds of physical versions): **0 terminal→non-terminal regressions** — the pre-fix clobber signature is absent
- Derive guard fallbacks ("stale-clone protection is off"): **0** · panics/fatals: **0** · pod restarts: **0**
- API-vs-ClickHouse cross-checks: identical on every slip
- All `state_history_lost` warns = known-benign D3 suppression class, columns proven consistent

**Independent oracle:** every verdict came from direct ClickHouse SQL (top-row columns vs `argMax` over `slip_component_states`) + API cross-check — not from the pipeline's own reporting.

### Causal differential: regression suite RED on v1.3.95, GREEN on this branch

The 8 testcontainers regression tests were run unmodified against baseline `main` (`b80da899`, v1.3.95) on the same live ClickHouse 25.8 — the only variable is this branch's code:

| Test | v1.3.95 baseline | This branch |
|---|---|---|
| `StatusPath_HealsStaleSiblingStep` (74cd6676 interleaving) | **FAIL** — `dev_deploy_status = "running", want "completed"` | PASS |
| `HistoryPath_HealsStaleSiblingStep` | **FAIL** — same stale re-seal | PASS |
| `TerminalSeal_EndToEndThenStable` | **FAIL** — stale value sealed on terminal slip | PASS |
| `ParityWithDerivePipelineStepStatuses` | **FAIL** — client-side derive `completed` vs SQL-side clone `pending` | PASS |
| `CancelBranch_MirrorsPreUpdateRowVerbatim` | FAIL (at its heal-precondition sanity gate) | PASS |
| Override precedence · no-events fallback · aggregate verbatim | PASS | PASS |

The three both-PASS rows prove the change is surgical: pre-existing semantics (tier-1 override, verbatim fallback, aggregate handling) are unchanged; only the stale-clone behavior differs. Combined with the prod forensics (corr `74cd6676`, the live wedge on v1.3.95), the proof chain is causal: same interleaving → fails on old code → heals on this code.

**Live A/B confirmation (same day):** the test environment was temporarily reverted to v1.3.95 and the identical storm replayed. The baseline binary produced a live clone-born stale regression under ordinary concurrent traffic — `secret_scan_status` `completed → running → completed` across row versions ~190 ms apart, with no backing event in `slip_component_states` (provably written by a concurrent verbatim clone from a stale snapshot; it survived only because a fresher write happened to follow). The patched build ran roughly twice that load across two storms plus failure-injection with zero such regressions in any version walk. Environment restored to the patched build afterward.

---


pkuzmenko-mycarrier:
**Concurrent-storm validation: PASSED**

Ran a multi-scenario concurrent-storm validation against `slippy-api-test` (image `26.29.70db41e`, ClickHouse `ci_test`) embedding this fix.

- 9 slips judged in the final full-window sweep, including 8 terminal (failed/abandoned) wedge-prone slips
- Scenarios covered: retry/supersede races at 3 depths, same-SHA retry reuse, a double `unit_tests` handler race (74cd6676 analog), late-completion heals on abandoned slips, and offload/auto-tests concurrency
- **0** column-vs-event mismatches, **0** derive-guard fallbacks (`stale-clone protection is off`), **0** pod restarts

Full report: `standup-notes/2026/07/slippy-5dv5-validation/verdict.md`

slippy/clickhouse_store.go
```diff
@@ -1448,6 +1449,257 @@ func (s *ClickHouseStore) appendHistoryWithOverrides(
...
+		// "HAVING st != ''" clause to mirror the Go guard, but comparing an Enum8
+		// column against the string literal '' is itself invalid for this enum type —
+		// ClickHouse rejects it at query-analysis time with error 691 ("Unknown
+		// element '' for enum"), failing the query outright (even with zero matching
+		// rows), rather than filtering as intended. Bd mycarrier-5dv5.
+		withClause = fmt.Sprintf(
+			"WITH derived AS (SELECT step, argMax(status, %s) AS st FROM %s.%s"+
+				" WHERE correlation_id = ? AND component = '' GROUP BY step) ",
+			componentEventSortKeyNoImageTag, database, TableSlipComponentStates,
+		)
```
smunukutla-mycarrier, line 1672: **Merge-gate coverage gap on the CLONE_DERIVED derive SQL** *(Logic, important)*

The 3-tier precedence SQL introduced here — the `WITH derived AS (… argMax(status) … GROUP BY step)` CTE, the per-step `coalesce((SELECT st FROM derived WHERE step='…'), col)`, and the conditional CTE-arg binding in both writers — depends on runtime-DB behaviours that only a live ClickHouse exercises: the tier-2→tier-3 NULL-fallthrough, argMax precedence, and the `component=''`/`GROUP BY` shape. You validated these 8/8 on live CH 25.8, so this isn't an open correctness question — it's a **durable coverage gap**. The tests asserting these (`clickhouse_store_stale_clone_regression_integration_test.go`, `clickhouse_store_i5_v2_integration_test.go`) carry `//go:build integration`, and neither the merge gate (`.github/workflows/ci.yml:70`) nor the root `Makefile` `test` target passes `-tags integration`; no integration-CI job exists. A future edit that breaks the derive query text, the arg binding, or the Enum8 handling therefore passes CI green.

The blast radius is asymmetric, and that's what makes this *important* rather than a suggestion: on the status path (`checkPipelineCompletion` → `updateSlipStatusWithOverrides` → `insertAtomicStatusUpdate`) the error **propagates** (`ErrSlipStatusUpdateFailed`) and, outside shadow mode, blocks every slip-status transition. On the audit path (`AppendHistory`/`UpdateStepWithHistory` → `insertAtomicHistoryUpdate`) the same error is swallowed best-effort and self-heals on next Load.

**Recommendation:** add a merge-gated regression that exercises the derive SQL end-to-end — a CI job/matrix entry running `go test -tags integration ./slippy/...` against an ephemeral ClickHouse (testcontainers or a service container), or promote a minimal NULL-fallthrough + tier-precedence assertion into an untagged test. The tier-2→tier-3 coalesce fallthrough and the conditional-CTE arg binding should be covered by whatever suite gates merges.

slippy/clickhouse_store.go
```diff
@@ -1448,6 +1449,257 @@ func (s *ClickHouseStore) appendHistoryWithOverrides(
...
+				reason: cloneDeriveFallbackReasonNonIdentifierGuard,
+			})
+			continue
+		}
+
+		needsDerive = true
+		exprs = append(exprs, fmt.Sprintf(
+			"coalesce((SELECT st FROM derived WHERE step = '%s'), %s)",
+			sqlSingleQuoteEscape(stepName), col,
+		))
```
smunukutla-mycarrier, line 1651: **tier-2 argMax can transiently revert a freshly-overridden sibling column** *(Concurrency, suggestion)*

For a non-overridden pure step, this `coalesce` picks the tier-2 argMax event value over the tier-3 cloned column value whenever *any* event exists for that step. That opens an inversion window: if a concurrent writer set a sibling column via a tier-1 override, and that override's `routing_slips` row becomes visible to this clone SELECT *before* its backing `slip_component_states` event lands, this write re-derives the sibling from the older visible event and reverts the fresher override. On a terminal slip the reverted value persists in the snapshot.

This stays a suggestion (not a defect) for two reasons: (1) the net is **positive** — `UpdateStepWithHistory` writes the event first (`insertComponentState`) then the row (`appendHistoryWithOverrides`), so "row visible before its own event" (the window this opens) is strictly rarer than "event visible before its row" (the window this fix closes); (2) the blast radius is **identical to the defect being fixed** — both corrupt only the `routing_slips` snapshot column, which `hydrateSlip` recomputes from the `component=''` event log on every Load (invariant #10), so both self-heal on the next Load. Consider noting the residual inversion window in the function doc so the "re-seal permanently" framing isn't read as fully closed.

slippy/clickhouse_store_clone_derive_test.go
```diff
@@ -0,0 +1,622 @@
...
+		// [0] correlationID(cancel), [1] newVersion(cancel), [2] correlationID(CTE),
+		// [3] newStatus, [4] newVersion(new-row), [5] override value, [6] correlationID(final)
+		if len(args) != 7 {
+			t.Fatalf("attempt %d: expected 7 args, got %d: %v", i, len(args), args)
+		}
+		if args[5] != string(StepStatusFailed) {
+			t.Errorf("attempt %d: expected override arg at index 5 to be %q, got %v",
+				i, string(StepStatusFailed), args[5])
+		}
+	}
```
smunukutla-mycarrier, line 500: **Status-writer arg-order test pins only length and index-5, so a positional swap can pass CI** *(test robustness, suggestion)*

`insertAtomicStatusUpdate` and `insertAtomicHistoryUpdate` assemble `execArgs` in the same lockstep order (cancel `corrID`+`version`, optional CTE `corrID`, literal(s), override args, final `corrID`), and a drift in one isn't caught by the other. The history writer is fully protected — `TestInsertAtomicHistoryUpdate_ArgOrdering_OverrideAndDerive` value-pins all seven positions. This status-writer test is the only check on `insertAtomicStatusUpdate`'s arg vector, and it asserts only `len(args)==7` and `args[5]`. A positional swap preserving the count and index-5 (e.g. transposing the `newStatus` literal and the override) would misbind parameters and still pass CI.

Subtlety in the fix: `SlipStatusFailed` and `StepStatusFailed` **both stringify to `"failed"`**, so pinning positions isn't enough unless `newStatus` differs from the override string — otherwise a `[3]↔[5]` swap (the exact bug class this guards against) stays invisible. The suggestion passes `SlipStatusCompleted` as `newStatus` (distinct from the `"failed"` override) and pins all seven positions:

```suggestion
	override := stepStatusOverride{columnName: "push_parsed_status", status: StepStatusFailed}
	err := store.insertAtomicStatusUpdate(
		// newStatus ("completed") is deliberately distinct from the override status
		// ("failed") so a [3]<->[5] transposition can't hide behind identical strings.
		context.Background(), "corr-ancestry-retry", uint64(7), SlipStatusCompleted, override,
	)
	if err != nil {
		t.Fatalf("expected no error, got %v", err)
	}

	if len(stmts) != 2 {
		t.Fatalf("expected exactly 2 ExecWithArgs calls (initial + ancestry retry), got %d", len(stmts))
	}
	if !strings.Contains(stmts[0], ColumnAncestry) {
		t.Error("expected first attempt to include the ancestry column")
	}
	if strings.Contains(stmts[1], ColumnAncestry) {
		t.Error("expected retry to omit the ancestry column")
	}

	// Both attempts must still carry the CLONE_DERIVED CTE (dev_deploy is pure/unoverridden)
	// and the override value, in the same relative argument order.
	for i, stmt := range stmts {
		if !strings.Contains(stmt, "WITH derived AS") {
			t.Errorf("attempt %d: expected CLONE_DERIVED CTE to survive the ancestry retry, stmt: %s", i, stmt)
		}
	}
	want := []interface{}{
		"corr-ancestry-retry",       // [0] correlationID (cancel WHERE)
		uint64(7),                   // [1] newVersion (cancel WHERE)
		"corr-ancestry-retry",       // [2] correlationID (CTE WHERE)
		string(SlipStatusCompleted), // [3] newStatus literal
		uint64(7),                   // [4] newVersion (new-row literal)
		string(StepStatusFailed),    // [5] push_parsed_status override
		"corr-ancestry-retry",       // [6] correlationID (final WHERE)
	}
	for i, args := range argSets {
		if len(args) != len(want) {
			t.Fatalf("attempt %d: expected %d args, got %d: %v", i, len(want), len(args), args)
		}
		for j, w := range want {
			if args[j] != w {
				t.Errorf("attempt %d: args[%d] = %v, want %v", i, j, args[j], w)
			}
		}
	}
```

slippy/push.go
```diff
@@ -719,14 +719,29 @@ func (c *Client) handlePushRetry(ctx context.Context, slip *Slip) (*Slip, error)
...
+	// the retry.
+	if err := c.store.UpdateStepWithHistory(
+		ctx,
+		slip.CorrelationID,
+		"push_parsed",
+		"",
+		StepStatusRunning,
+		entry,
+	); err != nil {
+		return nil, fmt.Errorf("%w: retry push_parsed reset failed: %s", ErrHistoryAppendFailed, err.Error())
```
smunukutla-mycarrier: **`ErrHistoryAppendFailed` is inverted on the retry path** *(Error-Handling, suggestion)*

This wraps every error from `UpdateStepWithHistory` as `ErrHistoryAppendFailed`, but after the reroute that method returns an error **only** when the terminal-freshness gate rejects the write (`ErrTerminalAlreadyExists`) or the event insert (`insertComponentState`) fails — the actual `state_history` write-back is now best-effort and swallowed (`return nil`). So the sentinel is inverted relative to its doc ("failed to append state history", `errors.go:52`): a genuine history-append failure now reports success, while gate/event-insert failures are mislabeled as history-append failures. Pre-refactor the sentinel was attached precisely to a `c.store.AppendHistory` failure, which was accurate.

No caller does `errors.Is` on it today, so the impact is diagnostic — an operator debugging a retry failure sees "failed to append state history" when the real cause was a gate rejection or event-insert failure. Wrapping the underlying error with `%w` preserves the real cause (including `ErrTerminalAlreadyExists`, so `errors.Is` against the true error works):

```suggestion
		return nil, fmt.Errorf("retry push_parsed reset failed: %w", err)
```
(This leaves `ErrHistoryAppendFailed` defined-but-unused; it's exported, so the `unused` linter won't flag it. Drop the definition too, or repurpose it for the genuinely-swallowed write-back path.)

smunukutla-mycarrier (changes requested):

slippy/clickhouse_store.go
pkuzmenko-mycarrier: Fixed in 2ed8be8 — you're right that the derive SQL's load-bearing behaviors only exist under a real ClickHouse, and nothing in the merge gate ran them. Added an `integration-test-slippy` CI job that runs the deterministic CLONE_DERIVED regression suite (`-tags integration -run 'TestStaleCloneDerive'`, testcontainers CH 25.8) on the same triggers as unit tests — it covers exactly the behaviors you listed: tier-2→tier-3 NULL fallthrough, argMax precedence, the `component=''`/GROUP BY shape, and the conditional CTE arg binding, on both writers.

One deliberate scoping call: the job gates the 8 deterministic regression tests rather than the full `-tags integration` suite, because the full suite carries pre-existing flakes unrelated to this change (`TestVersionedCollapsingMergeTree_DataIntegrity`, `TestHydrateSlip_EmptyAggregates`, `C_RecoveryRace` — the last tracked as bd `mycarrier-m1uz`). Gating those today would redline CI on day one; broadening the gate belongs with stabilizing them. Also flagging: making the new job a required branch-protection check is a repo-settings step outside ci.yml — happy to follow up if you want it enforced there too.

slippy/clickhouse_store.go
pkuzmenko-mycarrier: Doc note added in 2ed8be8, per your suggestion — `buildCloneStepColumnDerive`'s doc block now names the residual tier-2 inversion window explicitly, so nobody reads the fix as closing every interleaving. Your analysis matches ours: the window needs the sibling's row to become visible before its own backing event, which inverts the write order every handler uses (event first, then row), so it's strictly rarer than the window this closes; and because gates and pipeline flow read `slip_component_states` rather than the snapshot column, the blast radius stays cosmetic-and-self-healing. We saw the ordering hold empirically across two live storms plus an A/B replay against v1.3.95 — zero tier-2 inversions observed, while the baseline binary produced a genuine clone-born regression under the same load (details in the PR description's A/B section).

slippy/clickhouse_store_clone_derive_test.go
pkuzmenko-mycarrier: Fixed in 2ed8be8 — the status-writer test now value-pins all seven positions (cancel corrID, cancel version, CTE corrID, status literal, version literal, override arg, final corrID), mirroring the history-writer test's pattern, with the expected order verified against `insertAtomicStatusUpdate`'s own assembly. A transposition of any two positions now fails the test rather than sliding through on `len==7`. Good catch — the two writers assemble their arg vectors in lockstep and deserve symmetric protection.

slippy/push.go
pkuzmenko-mycarrier: Fixed in 2ed8be8 — sharpest catch of the round: the sentinel was accurate before the reroute and we inverted it without noticing. `handlePushRetry` now wraps the real underlying error (`retry push_parsed reset failed: %w`), so `errors.Is(err, ErrTerminalAlreadyExists)` works through the chain and an operator sees the true cause (gate rejection or event-insert failure) instead of a misleading 'failed to append state history'. Verified `push.go:744` was the only site returning `ErrHistoryAppendFailed` and nothing does `errors.Is` on it — the sentinel stays defined (public API) with its doc updated to reflect it's no longer produced here.

smunukutla-mycarrier (approved):

PR #80, opened 2026-07-22
feat: Postgres storage foundation for Slippy — postgres, postgresmigrator, PostgresStore (DEVOPS-127)

smunukutla-mycarrier:
## Summary

Adds the **goLibMyCarrier foundation for migrating Slippy's operational slip storage from ClickHouse to Postgres** ([DEVOPS-127](https://linear.app/mycarrier/issue/DEVOPS-127)). Two new modules (`postgres`, `postgresmigrator`) plus a Postgres `SlipStore` implementation and schema generator inside `slippy`.

This PR is **purely additive and inert**: it introduces new code paths but does **not** wire them into any running service and does **not** change the existing ClickHouse store. Consumers (slippy-api read/write split, the migrator Job, the data backfill) land in follow-up PRs. Nothing here executes until something constructs a `PostgresStore`.

> Full design + old-vs-new DB mapping are attached to the Linear ticket (v4 docs).

## Why

ClickHouse has no atomic in-place update, so every slip "update" is an insert that an async background merge later collapses. Between write and merge, reads can see stale/uncollapsed rows — which is why both `goLibMyCarrier/slippy` and slippy-api carry a large stack of consistency workarounds (verification-retry loops, the I5 freshness gate, write-fingerprint suppression, clone-derive, BC-13 mutual-invisibility, read-your-writes overlays).

Postgres (row store + MVCC + transactions) gives atomic, immediately-visible `UPDATE … WHERE`. **The reviewer takeaway: `PostgresStore` is dramatically smaller than `clickhouse_store.go`, and that's the point** — nearly all of that machinery is async-merge tax that simply doesn't exist under MVCC. What remains is pure domain logic.

## What's in this PR

Each commit is a self-contained, green module (TDD; unit + integration):

| Commit | Module | What |
|---|---|---|
| `1838239`, `597df7e` | **`postgres`** (new) | pgx-based client lib mirroring `clickhouse`: `POSTGRES_*` config loader/validator + `pgxpool` session (connect/ping/close, retry, TLS, pool tuning). |
| `7a24d01` | **`postgresmigrator`** (new) | Versioned migrations + idempotent ensurers over pgx, mirroring `clickhousemigrator` — but each migration applies **transactionally** (Postgres has transactional DDL). |
| `a4f23e4` | `slippy` | Postgres dynamic-schema generator + `RunPostgresMigrations`. Greenfield v1 schema: `slip_status`/`step_status` **DOMAINs** (Enum8 → text+CHECK), `jsonb`, `timestamptz`, **no sign/version**, current-state `slip_component_states` keyed by `(correlation_id, step, component)`. Per-step status + aggregate columns emitted as config-driven ensurers, exactly like the CH generator. |
| `df6d7cb` | `slippy` | `PostgresStore` **2a** — core CRUD (`Create`/`Update`/`Load`/`LoadByCommit`/`LoadLiveByCommit`/`Close`/`Ping`) + row hydration. |
| `0168e72` | `slippy` | `PostgresStore` **2b** — `FindByCommits`/`FindAllByCommits` (via `unnest … WITH ORDINALITY`) + ancestry (`InsertAncestryLink`/`ResolveAncestry`). |
| `ab856fb` | `slippy` | `PostgresStore` **2c** — `UpdateStep`/`UpdateComponentStatus`/`UpdateStepWithHistory`/`AppendHistory`/`UpdateSlipStatus`/`SetComponentImageTag` + in-transaction aggregate recompute. Completes the `SlipStore` interface (compile-time asserted). |
| `701b47a` | (several) | Routine `go.mod`/`go.sum` dependency bumps + `makefile` `LIB_DIRS` registration for the two new modules. |

`PostgresStore` implements the existing **backend-agnostic `SlipStore` interface** — the state machine (which lives in `slippy.Client`, not the store) is untouched.

## Design notes worth a reviewer's eye

- **Terminal-monotonicity is preserved** (prod behavior), re-expressed transactionally: a `WHERE`-clause guard on the component upsert rejects a non-terminal status overwriting a terminal one for the same step (with the `aborted → pending` retry-reset exception), returning `ErrTerminalAlreadyExists`. No timing window / `SLIPPY_I5_*` knobs — that was ClickHouse-specific.
- **One transaction per mutation, serialized per slip** via `SELECT … FOR UPDATE`. The component upsert + aggregate recompute + `state_history` append + step-status column all commit atomically — so, unlike the CH best-effort write-back, a history audit entry is never silently dropped.
- **`slip_component_states` collapses to a current-state table** (PK upsert) rather than an append-only event log; validated by the PoC. Aggregate status is recomputed over the active components (`computeAggregateStatus` precedence + placeholder exclusion carried over verbatim).
- **JSON shape parity with ClickHouse** for `step_details` / `state_history` (`{"entries":[…]}`) / aggregate (`{"items":[…]}`) columns — deliberate, so the one-time data backfill and the ClickHouse-over-Postgres reporting reads stay byte-compatible.
- `LoadLiveByCommit` keeps the `{abandoned, promoted, compensated}` exclusion + case-insensitive repo match; `SlipStatus.IsTerminal` still treats `failed` as non-terminal.

## Blast radius

- **No existing module's behavior changes** — the only edits outside the new files are dependency-version bumps and the `LIB_DIRS` line. `clickhouse_store.go` is untouched.
- Nothing is imported by a running service in this PR.

## Testing

- **Unit**: table-driven; the store is exercised against **`pgxmock`** (no DB) so the SQL/branching is covered in the normal `make test` run.
- **Integration**: `//go:build integration` + Postgres **testcontainers**, matching the existing ClickHouse integration gating (excluded from the default build; run in the Docker CI job). Covers migrate/validate/idempotency, CRUD round-trips, commit-priority + status exclusion, cross-branch ancestry + `maxDepth`, the aggregate rollup, `StartedAt` preservation across transitions, the terminal guard (block + `aborted→pending`), and history ordering.
- Coverage is above the CI per-module gate (postgresmigrator ~83%, slippy ~83%).

## Suggested review order

1. `postgres/` (config + session) — smallest, self-contained.
2. `postgresmigrator/` (migrator.go) — note the transactional `inTx` per-migration application.
3. `slippy/postgres_migrations.go` + `postgres_migrate.go` — the generated schema + `RunPostgresMigrations`.
4. `slippy/postgres_store.go` → `postgres_store_reads.go` → `postgres_store_updates.go` — the store, in interface order.

## Out of scope (follow-ups, not in this PR)

- slippy-api read/write split (command path → `PostgresStore`, query path → ClickHouse-over-Postgres reader) — separate slippy-api PR.
- The schema-only migrator k8s Job binary — separate slippy-api PR (already built).
- The standalone one-time `copy-data` backfill script (run at cutover).
- The ClickHouse-over-Postgres reporting bridge + `pg_cron`/partitioning retention.


slippy/postgres_store_updates.go
```diff
@@ -0,0 +1,399 @@
...
+		terminalStepStatusesSQL, nonTerminalStepStatusesSQL)
+
+	tag, err := tx.Exec(ctx, q, correlationID, step, component, string(status), message, imageTag)
+	if err != nil {
+		return false, fmt.Errorf("failed to record component state %s/%s: %w", step, component, err)
+	}
+	// A pure INSERT or a guard-passing DO UPDATE affects one row; a guard-blocked conflict
+	// affects zero.
+	return tag.RowsAffected() > 0, nil
+}
```
pkuzmenko-mycarrier, line 236: **Terminal guard is permanent — it drops two real ClickHouse invariants, not just async-merge tax.**
`upsertComponentState` rejects any non-terminal status overwriting a terminal one for the same step (except `aborted→pending`), unconditionally and forever. The ClickHouse gate this replaces (`enforceTerminalFreshnessGate`, `clickhouse_store.go:2728-2812`) is a **~750ms freshness window** with two escape hatches that are documented as intentional (`clickhouse_store.go:86-89`: "push-webhook retries legitimately reset a terminal push_parsed event to running… safe-by-design, not a race artifact"):
1. **age > window ⇒ allow** — a genuine re-run/restart is permitted once the prior event is older than the window.
2. **`gateBypassSteps`** exempts pipeline-level `push_parsed` writes.

Concrete break once this store is wired: `handlePushRetry` (`push.go:742-751`) issues `UpdateStepWithHistory(corr, "push_parsed", "", running)` while the slip is live but `push_parsed` is already `completed`. Under this guard the upsert affects 0 rows → `ErrTerminalAlreadyExists` → `CreateSlipForPush` fails on an ordinary re-push. The same rejects a `StartStep` re-run against a terminal step (the documented pre-job "StartStep against a terminal slip" flow).

MVCC removes the async-insert duplicate-write race (the sub-window refusal), but it does **not** provide the time-based re-run discrimination the gate exists for. Please either port the freshness window + `push_parsed` bypass (time-bound the guard: `OR (now() - slip_component_states.updated_at) > $n::interval`, and skip the guard for bypass steps), or get explicit sign-off that Postgres deliberately forbids all terminal→non-terminal re-runs and adjust the callers accordingly.

slippy/postgres_store_updates.go
```diff
@@ -0,0 +1,399 @@
...
+		return fmt.Errorf("failed to marshal aggregate %s: %w", aggStep, err)
+	}
+
+	upd := fmt.Sprintf(
+		"UPDATE routing_slips SET %s_status = $1, %s = $2, updated_at = now() WHERE correlation_id = $3",
+		aggStep, aggStep)
+	if _, err := tx.Exec(ctx, upd, string(status), string(itemsJSON), correlationID); err != nil {
+		return fmt.Errorf("failed to write aggregate %s: %w", aggStep, err)
+	}
+	return nil
```
pkuzmenko-mycarrier, line 319: **Empty active-component set writes `completed`, inverting reality.** `pgComputeAggregateStatus([])` returns `StepStatusCompleted` (the loop never runs, `allCompleted` stays `true`), and `recomputeAggregate` writes the result unconditionally. Reached via `UpdateStep(corr, <aggregateStep>, "", status)` before any component rows exist: the just-inserted `component=''` row is excluded by the `component <> ''` filter, so `active` is empty and `{aggStep}_status` is set to `completed` with zero components reporting. ClickHouse deliberately guards this (`if len(active) > 0` / `if !updated { return }`, `clickhouse_store.go:2979,3491`, comment about `builds_status` getting stuck). A downstream deploy/prereq gate reading that `completed` would mis-gate.
```suggestion
	// Leave the aggregate untouched when there are no active components — matches
	// ClickHouse's `if len(active) > 0` guard and avoids forcing an empty aggregate to 'completed'.
	if len(active) == 0 {
		return nil
	}
	status := pgComputeAggregateStatus(active)
```

slippy/postgres_store.go
```diff
@@ -0,0 +1,419 @@
...
+
+	tag, err := s.pool.Exec(ctx, query, args...)
+	if err != nil {
+		return fmt.Errorf("failed to update slip %s: %w", slip.CorrelationID, err)
+	}
+	if tag.RowsAffected() == 0 {
+		return ErrSlipNotFound
+	}
+	return nil
+}
```
pkuzmenko-mycarrier, line 141: **`Update` is an unlocked full-row read-modify-write — lost updates are possible and permanent.** Unlike every other mutator (which wraps `lockSlip`/`SELECT … FOR UPDATE` in `inTx`), `Update` (`postgres_store.go:99-131`) runs a single `UPDATE … SET <all columns> WHERE correlation_id` from a caller-supplied `*Slip` snapshot with no lock. `AbandonSlip`/`PromoteSlip` (`client.go:169-219`) `Load` (unlocked) then `Update`. A concurrent `updateStepTx` holding `FOR UPDATE` that commits between the `Load` and the `Update` is clobbered: all `*_status` columns, aggregate JSONB, and `state_history` are overwritten from the stale snapshot (a wholesale overwrite, not the `jsonb_set` append the history path uses — so concurrent audit entries are silently lost). And since `populate` (`218-254`) reads columns directly and never re-derives from `slip_component_states`, the stale value is permanent until the next component update recomputes that aggregate. The CH "unique nanosecond timestamp ⇒ no conflicts" property (`client.go:180-181`) does not carry over — CH re-derived on Load. Route `Update` through `inTx + lockSlip`, or scope it to only the columns abandon/promote actually change.

slippy/postgres_store_updates.go
```diff
@@ -0,0 +1,399 @@
...
+		"SELECT correlation_id FROM routing_slips WHERE correlation_id = $1 FOR UPDATE",
+		correlationID).Scan(&id)
+	if err != nil {
+		if isNoRows(err) {
+			return ErrSlipNotFound
+		}
+		return fmt.Errorf("failed to lock slip %s: %w", correlationID, err)
+	}
+	return nil
+}
```
pkuzmenko-mycarrier, line 396: **`UpdateStep` on a not-yet-created slip loses the event instead of absorbing the ordering.** `lockSlip` requires the `routing_slips` row to exist and returns `ErrSlipNotFound` immediately, rolling back the whole tx with `upsertComponentState` never reached — the component event is **not recorded**. ClickHouse instead (a) records the step event first, durable even with no slip row (`clickhouse_store.go:659-706`), and (b) retries slip-creation with linear backoff up to ~30 min (`2915-2947`), absorbing the documented producer/consumer race where a CI agent's `StartStep` (via slippy-api) arrives before pushhookparser's `Create`. Synchronous committed `Create` makes this far less likely, so a simplification may be fine — but dropping event-log durability + backoff is a real contract change. Please make it explicit: either decouple the component-event record from the `FOR UPDATE` (record without requiring the slip row), or add slip-not-found backoff and document a strict Create-before-UpdateStep ordering requirement for all callers.

postgres/postgres.go
```diff
@@ -0,0 +1,165 @@
...
+// Connection-pool defaults, applied when PostgresConfig leaves a field
+// zero-valued. ConnMaxLifetime is kept short so connections behind
+// LBs/firewalls that silently drop idle peers are reaped promptly.
+const (
+	DefaultMaxConns        int32         = 50
+	DefaultMinConns        int32         = 2
+	DefaultConnMaxLifetime time.Duration = 60 * time.Second
+
+	defaultPort    = "5432"
+	defaultSSLMode = "require" // Azure Database for PostgreSQL requires TLS by default.
```
pkuzmenko-mycarrier: **`sslmode` defaults to `require`, which does not protect against MITM.** Per PostgreSQL libpq (Table 32.1), `require` encrypts but performs no CA/hostname verification; `verify-full` is recommended in security-sensitive environments. `validSSLModes` (`postgres.go:31-33`) also admits `disable`/`allow`/`prefer`, so a mis-set `POSTGRES_SSLMODE` silently runs with no/opportunistic TLS and `PostgresValidateConfig` accepts it without warning. Against the Azure PG target, `verify-full` works out of the box via the system cert pool (DigiCert chain). Recommend defaulting to `verify-full` and restricting the accepted set for production — **but** ship it with documented `sslrootcert` guidance and an explicit override, otherwise consumers whose CA isn't in the system pool will fail closed on upgrade.

slippy/postgres_store_updates.go
```diff
@@ -0,0 +1,399 @@
...
+		if rbErr := tx.Rollback(ctx); rbErr != nil && !errors.Is(rbErr, pgx.ErrTxClosed) {
+			return errors.Join(err, fmt.Errorf("rollback: %w", rbErr))
+		}
+		return err
+	}
+	if err := tx.Commit(ctx); err != nil {
+		return fmt.Errorf("commit transaction: %w", err)
+	}
+	return nil
+}
```
pkuzmenko-mycarrier: Both `inTx` helpers roll back only on the explicit `fn(tx) != nil` path. A panic inside `fn` (or a future early return) skips both `Commit` and `Rollback`, so the pgx Tx never releases its pooled connection — repeated occurrences exhaust the pool. Add `defer func(){ _ = tx.Rollback(ctx) }()` right after `Begin`; a post-commit rollback is a safe no-op (`ErrTxClosed`, already tolerated here).

_(Same pattern also present in `postgresmigrator/migrator.go:436-451` — apply the same fix there.)_

postgresmigrator/migrator.go
```diff
@@ -0,0 +1,476 @@
...
+		return 0, fmt.Errorf("failed to check if %s exists: %w", qualified, err)
+	}
+	if !exists {
+		return 0, nil
+	}
+
+	var version int
+	err := m.conn.QueryRow(ctx,
+		fmt.Sprintf("SELECT version FROM %s ORDER BY applied_at DESC LIMIT 1", qualified),
+	).Scan(&version)
```
pkuzmenko-mycarrier, line 211: Current version is read via `ORDER BY applied_at DESC LIMIT 1`, and `applied_at = now()` (transaction-start time). Same-microsecond ties are unordered, and a backward NTP step could surface a stale row; the down path writes `previousVersion` with a *newer* `applied_at`, so ordering by version wouldn't be right for downgrades either. Add a monotonic `bigserial id` and order by `id DESC` — unambiguous for both up and down transitions. (Low realistic probability given infrequent single-writer migrations.)

postgres/go.mod
```diff
@@ -0,0 +1,79 @@
...
+	github.com/tklauser/go-sysconf v0.3.16 // indirect
+	github.com/tklauser/numcpus v0.11.0 // indirect
+	github.com/yusufpapurcu/wmi v1.2.4 // indirect
+	go.opentelemetry.io/auto/sdk v1.2.1 // indirect
+	go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.60.0 // indirect
+	go.opentelemetry.io/otel v1.41.0 // indirect
+	go.opentelemetry.io/otel/metric v1.41.0 // indirect
+	go.opentelemetry.io/otel/trace v1.41.0 // indirect
+	go.yaml.in/yaml/v3 v3.0.4 // indirect
+	golang.org/x/crypto v0.51.0 // indirect
```
pkuzmenko-mycarrier: `postgres` and `postgresmigrator` pin `golang.org/x/crypto v0.51.0` while the sibling `slippy` module already ships `v0.54.0`. Trivy flags a CVE cluster on 0.51.0 (fixed in 0.52.0); `govulncheck ./...` reports the vulnerable symbols (the `x/crypto/ssh` cluster) as **not reachable** — pgx only pulls `x/crypto/pbkdf2` for SCRAM — so this is dependency hygiene, not an active exploit. Bump both new modules to `v0.54.0` to align the monorepo floor and clear Trivy/Dependabot noise. (Scanner artifacts from this review: Semgrep 0 findings, govulncheck clean, Trivy 0.72.0.)

_(Same pin in `postgresmigrator/go.mod:67` — bump both.)_

postgres/postgres.go
```diff
@@ -0,0 +1,165 @@
...
+)
```
pkuzmenko-mycarrier, line 44: The pool sets sizing but no server-side `statement_timeout`, `lock_timeout`, or `idle_in_transaction_session_timeout`. Because `lockSlip` takes `SELECT … FOR UPDATE` inside a transaction, a hung/idle transaction holds the per-`correlation_id` row lock and blocks all other mutators for that slip until the pool lifetime or a client timeout intervenes. Set conservative server-side timeouts via the connection `RuntimeParams` (config-overridable) to bound runaway queries and prevent a stuck transaction from holding slip locks — cheap liveness/DoS hardening for a shared pool.

_(Also applies to the pool wiring in `postgres/session.go:51-58`.)_

pkuzmenko-mycarrier (changes requested):

slippy/postgres_store_updates.go
```diff
@@ -0,0 +1,399 @@
...
+				}
+			}
+		} else {
+			// Pure pipeline step: the status column on routing_slips is authoritative.
+			col := stepName + "_status"
+			upd := fmt.Sprintf("UPDATE routing_slips SET %s = $1, updated_at = now() WHERE correlation_id = $2", col)
+			if _, err := tx.Exec(ctx, upd, string(status), correlationID); err != nil {
+				return fmt.Errorf("failed to update step %s: %w", stepName, err)
+			}
+		}
```
bcarlock-mycarrier: **SQL identifier injection via unvalidated `stepName` (+ a correctness bug from the same gap).**

The pure-pipeline-step branch builds the target column from the caller-supplied `stepName` and interpolates it into the UPDATE with no validation:
```go
col := stepName + "_status"
upd := fmt.Sprintf("UPDATE routing_slips SET %s = $1, updated_at = now() WHERE correlation_id = $2", col)
```
Unlike the aggregate/config-derived paths, this uses the raw argument. `stepName` is attacker-influenceable end-to-end: slippy-api binds `StepName` from an HTTP path parameter with no `pattern`/`enum` (`slip_write_handler.go:87,102,112`) and passes it through `writer.*Step` -> `Client.UpdateStepWithStatus` -> `store.UpdateStepWithHistory`; `slip_component_states.step` is unconstrained `text`, so the bound upsert accepts any payload and execution reaches the interpolation. pgx's bound args block stacked queries, but the SET left-hand side is fully attacker-controlled — e.g. `stepName = "state_history = '[]'::jsonb, dev_deploy"` renders `UPDATE routing_slips SET state_history = '[]'::jsonb, dev_deploy_status = $1, updated_at = now() WHERE correlation_id = $2`, wiping the audit trail and setting an arbitrary status on the target slip.

Threat model: all slippy-api write routes are auth-gated (`writeApiKeySecurity`) and `lockSlip` requires the slip to already exist, so the realistic actor is an authenticated write client with a known `correlation_id` (a leaked write key or a malicious/compromised CI workflow definition) — squarely within this platform's threat model. It is latent until `PostgresStore` is wired as the `SlipStore`, but this PR introduces the code and the library's ClickHouse store already treats step names as an injection surface (`safeStepNameForDerivePattern = ^[A-Za-z0-9_]+$`, `clickhouse_store.go:1571`).

The same missing check is also a correctness bug independent of the attack: an unknown-but-safe step name (e.g. `foo`) targets nonexistent column `foo_status` -> pg error 42703, surfaced opaquely as *"failed to update step foo"*, and because the event upsert (:142) and this column UPDATE run in one transaction, it rolls back the already-written `slip_component_states` event row. ClickHouse tolerates unknown step names (records the event, materializes only config-known columns); Postgres hard-errors. One guard fixes both:

```suggestion
		} else {
			// Pure pipeline step: the status column on routing_slips is authoritative.
			// Guard the identifier interpolation: stepName reaches here from unvalidated
			// HTTP/CI input via the SlipStore interface, so require it to be (a) a known
			// configured step and (b) a bare identifier — reusing the library's own
			// safeStepNameForDerivePattern guard (clickhouse_store.go).
			if s.config.GetStep(stepName) == nil || !safeStepNameForDerivePattern.MatchString(stepName) {
				return fmt.Errorf("%w: invalid pipeline step %q", ErrInvalidConfiguration, stepName)
			}
			col := stepName + "_status"
			upd := fmt.Sprintf("UPDATE routing_slips SET %s = $1, updated_at = now() WHERE correlation_id = $2", col)
			if _, err := tx.Exec(ctx, upd, string(status), correlationID); err != nil {
				return fmt.Errorf("failed to update step %s: %w", stepName, err)
			}
		}
```

postgres/postgres.go
bcarlock-mycarrier: **`sslmode=require` encrypts but doesn't authenticate the server.** `defaultSSLMode = "require"` is applied whenever `POSTGRES_SSLMODE` is unset (:119) and is live-wired into the DSN (`session.go:37`). libpq/pgx `require` guarantees encryption but performs no CA or hostname verification, so an attacker with a network position between the service and the database can impersonate the Postgres server and feed forged routing-slip state to the CI/CD control plane. The hardened target posture for a control-plane datastore is `verify-full` (encrypt + validate the server cert chain and hostname).

Caveat before flipping the default: confirm the server CA is in the client trust store (system roots or an explicit `sslrootcert`), otherwise `verify-full` fails connections closed and causes an outage. `verify-ca` is a fallback where the presented hostname legitimately differs from the connection host.

```suggestion
	defaultSSLMode = "verify-full" // Target posture: encrypt AND verify server cert + hostname. Confirm the server CA is trusted (system roots or sslrootcert) before flipping, else connections fail closed.
```

postgres/go.mod
bcarlock-mycarrier: **Align `golang.org/x/crypto` with the sibling `slippy` module.** The two new modules pin `golang.org/x/crypto v0.51.0` (here and `postgresmigrator/go.mod:67`) while this PR's own `slippy/go.mod` is already at `v0.54.0` — an inconsistency the PR introduces. This is dependency hygiene, not a CVE exposure: `go list -deps` on the `postgres` module returns zero `x/crypto` packages, so the advisories that scanners match against `x/crypto/ssh` are definitively unreachable (a pgx Postgres client never imports the SSH package). In each new module run `go get golang.org/x/crypto@v0.54.0 && go mod tidy` to align with the already-patched sibling.

postgresmigrator/migrator.go
```diff
@@ -0,0 +1,476 @@
...
+// *pgxpool.Pool (e.g. goLibMyCarrier/postgres.PostgresSession).
+func NewMigratorFromSession(session PoolProvider, logger Logger, opts ...MigratorOption) (*Migrator, error) {
+	if session == nil {
+		return nil, ErrNilConnection
+	}
+	return NewMigrator(session.Pool(), logger, opts...)
+}
+
+// CreateTables applies pending migrations, then runs the ensurers.
+func (m *Migrator) CreateTables(ctx context.Context) error {
```
bcarlock-mycarrier, line 121: **No advisory lock serializes concurrent migration runs.** `CreateTables` reads the current version, then applies each pending migration and inserts a version row without holding a session-level lock across the run, so two concurrent migrator processes can both observe the same starting version and attempt the same migration. This is low-value hardening rather than a live risk: all current migrations are idempotent (`CREATE TABLE / ADD COLUMN / CREATE INDEX ... IF NOT EXISTS`), `GetSchemaVersion` uses `ORDER BY applied_at DESC LIMIT 1` so duplicate version rows are cosmetic, and `RunPostgresMigrations` is documented to run from a dedicated single Job. Residual value is hardening against a misconfigured multi-Job deployment or a future non-idempotent migration — wrap the run in `pg_advisory_lock`/`pg_try_advisory_lock` if concurrent migrators ever become possible.

slippy/postgres_store_updates.go
```diff
@@ -0,0 +1,399 @@
...
+		INSERT INTO slip_component_states (correlation_id, step, component, status, message, image_tag, updated_at)
+		VALUES ($1, $2, $3, $4, $5, $6, now())
+		ON CONFLICT (correlation_id, step, component) DO UPDATE SET
+			status     = EXCLUDED.status,
+			message    = EXCLUDED.message,
+			image_tag  = CASE WHEN EXCLUDED.image_tag <> '' THEN EXCLUDED.image_tag
+			                  ELSE slip_component_states.image_tag END,
+			updated_at = now()
+		WHERE NOT (slip_component_states.status IN (%s) AND EXCLUDED.status IN (%s))
+		   OR (slip_component_states.status = 'aborted' AND EXCLUDED.status = 'pending')`,
```
bcarlock-mycarrier: **This guard breaks step retries/re-runs — it is not a faithful re-expression of the ClickHouse terminal gate.** The `ON CONFLICT` WHERE guard (:194-195) rejects ANY non-terminal status overwriting a terminal one, with the single exception `aborted->pending`. The ClickHouse `enforceTerminalFreshnessGate` (`clickhouse_store.go:2728-2811`) refuses such a write *only* when the prior terminal event landed inside a freshness window (default 750ms, Rule 7 at :2807-2809) and explicitly ALLOWS it once the event is older ("genuine re-run / restart"). Postgres serialization/MVCC removes the *race-artifact* half of the gate's job, but Rule 7's legitimate-re-run allowance is required regardless of concurrency model — this port drops the time dimension entirely.

Concrete break: a step that reached `failed` (terminal) is retried by re-running its pre-job -> `StartStep -> UpdateStepWithStatus(running) -> UpdateStepWithHistory`. `failed->running` is non-terminal-over-terminal and isn't the `aborted->pending` exception, so `upsertComponentState` returns `applied=false` and `updateStepTx` returns `ErrTerminalAlreadyExists` (:146-149). The executor's reconcile only resets *cascade-aborted* steps to pending (`executor.go:376-380`); the primary `failed` step re-runs directly via `StartStep` and is permanently rejected. Under ClickHouse it succeeds after 750ms.

Second break: the ClickHouse gate bypasses the guard for pipeline-level `push_parsed` writes (`gateBypassSteps`, `clickhouse_store.go:90-92`) because push-webhook retries legitimately reset a terminal `push_parsed` to running. The Postgres guard has no bypass-step concept, so those resets are blocked too.

This is latent (nothing is wired yet) but is a contract divergence in the store whose entire purpose is to become the backend-agnostic `SlipStore` — it will surface as broken retries only after cutover. Fix requires reintroducing a time-based allowance (compare the existing row's `updated_at` against a freshness window, permit the write when the prior terminal event is older) plus a `push_parsed` bypass, rather than the unconditional WHERE clause.

slippy/postgres_store_updates.go
```diff
@@ -0,0 +1,399 @@
...
+			byName[row.Component] = len(items) - 1
+		}
+	}
+	if err := rows.Err(); err != nil {
+		return fmt.Errorf("failed to iterate component states for %s: %w", aggStep, err)
+	}
+
+	// Status is computed over the active components only (excludes any config placeholders
+	// left in the items list), matching filterActiveComponents in the ClickHouse store.
+	status := pgComputeAggregateStatus(active)
```
bcarlock-mycarrier: **An aggregate with no component rows yet is written as `completed`.** `pgComputeAggregateStatus(active)` returns `StepStatusCompleted` for an empty slice (`allCompleted` stays true through the empty loop — :369-398, locked in by the unit test at `postgres_store_updates_test.go:28`), and `recomputeAggregate` writes it unconditionally (:265, 274-279). This is reachable in normal operation: the `builds` aggregate emits a pipeline-level `running` event before any component reports — `StartStep(corr, "builds", "") -> UpdateStepWithHistory("builds", "", running)` — and `Create` seeds no placeholder component rows (`postgres_store.go:69-95`), so `active` is genuinely empty and `builds_status` is set to `completed`.

This diverges from ClickHouse, which the doc comment at :368 explicitly claims parity with: `computeAggregateStatus` is only ever invoked behind a `len(active) > 0` guard (`clickhouse_store.go:2979-2980, 3138-3139`) and `applyComponentStatesToAggregate` early-returns without touching status when there are no component states (:3491-3493). It never writes `completed` for an empty set. Beyond the wrong value, this has a gate-integrity dimension: a false `builds_status='completed'` reaches prerequisite evaluation (`getPrereqStatus` -> `CheckPrerequisites` treats `IsSuccess()` as satisfied, `prereqs.go:32,72-73`), so a downstream step gated on `builds` could advance prematurely. The wrong value self-corrects once the first component row lands, but is permanent for a build phase where no component ever reports.

```suggestion
	// Empty active set: no component rows exist for this aggregate yet. Match the
	// ClickHouse store (applyComponentStatesToAggregate early-returns) and leave the
	// existing aggregate status untouched rather than overwriting it with 'completed'.
	if len(active) == 0 {
		return nil
	}

	// Status is computed over the active components only (excludes any config placeholders
	// left in the items list), matching filterActiveComponents in the ClickHouse store.
	status := pgComputeAggregateStatus(active)
```

slippy/postgres_store_updates.go
```diff
@@ -0,0 +1,399 @@
...
+	case anyFailed:
+		return StepStatusFailed
+	case allCompleted:
+		return StepStatusCompleted
+	case anyRunning || anyCompleted:
+		return StepStatusRunning
+	default:
+		return StepStatusPending
+	}
+}
```
bcarlock-mycarrier: **Extract this instead of duplicating the ClickHouse aggregate-status logic.** `pgComputeAggregateStatus` reproduces the same `allCompleted/anyRunning/anyCompleted/anyFailed` accumulation and the same result precedence as `ClickHouseStore.computeAggregateStatus` (`clickhouse_store.go:2644-2675`). They're logically identical, not textually identical — ClickHouse uses a sequential if-chain and Postgres uses a `switch` — but the decision logic is the same. The ClickHouse method uses no receiver state, so the logic is already backend-agnostic, and both functions live in the same `slippy` package, so there's zero packaging barrier to extraction. The PR already shares `buildComponentData`, `updateExistingComponent`, and `reconstructStepTimingFromHistory`; the aggregate-status rule is the one piece of shared meaning that got copied. Any future change to the precedence (e.g. how `skipped`/`timeout` are weighed) must be made in two places or the backends silently diverge. Extract one free function (e.g. `computeAggregateStatus(components []ComponentStepData) StepStatus`) and have both stores call it.

bcarlock-mycarrier (changes requested):

slippy/postgres_store_updates.go
smunukutla-mycarrier: Addressed in d469c3e: ported the ClickHouse terminal-freshness gate into the upsert WHERE clause — reject terminal→non-terminal only within the freshness window, with the push_parsed bypass and the SLIPPY_I5_GATE_ENABLED toggle (same SLIPPY_I5_* knobs). Genuine re-runs (failed→running, push retries) are allowed once past the window. Age is computed on the Postgres server clock (skew-immune). Unit + integration tests added.

slippy/postgres_store_updates.go
smunukutla-mycarrier: Fixed in d469c3e: recomputeAggregate now returns early on len(active)==0, so an aggregate with no reporting components is left untouched instead of resolving to 'completed'. Test added.

slippy/postgres_store.go
smunukutla-mycarrier: Fixed in d469c3e: PostgresStore.Update now runs inside inTx+lockSlip, and AbandonSlip was moved off Load+Update onto the atomic UpdateSlipStatus (which also fixes the same lost-history race for ClickHouse). PromoteSlip is entangled with the unpersisted PromotedTo field, so it's tracked as a follow-up: DEVOPS-202.

slippy/postgres_store_updates.go
smunukutla-mycarrier: Addressed in d469c3e (decision: document the contract). Synchronous committed Create makes the ordering race rare, so lockSlip now documents the strict Create-before-UpdateStep contract: ErrSlipNotFound propagates and upstream Kafka at-least-once redelivery retries once Create lands, rather than the store buffering the event.

postgres/postgres.go
smunukutla-mycarrier: Fixed in cde00d4: defaultSSLMode is now verify-full, plus an optional POSTGRES_SSLROOTCERT override for a private CA / non-system trust store.

slippy/postgres_store_updates.go
smunukutla-mycarrier: Fixed in d469c3e (slippy) and 785389a (postgresmigrator): both inTx helpers now defer a rollback guarded by a committed flag, so a panic in fn always releases the transaction's connection.

postgresmigrator/migrator.go
smunukutla-mycarrier: Fixed in 785389a: schema_version is now ordered by a monotonic bigserial id instead of applied_at; the id is backfilled on pre-existing tables via ADD COLUMN IF NOT EXISTS, and the CREATE + backfill run in one transaction.

postgres/go.mod
smunukutla-mycarrier: Fixed in cde00d4 (postgres) and 785389a (postgresmigrator): bumped golang.org/x/crypto to v0.54.0 to align with the slippy module.

postgres/postgres.go
smunukutla-mycarrier: Fixed in cde00d4: added conservative, env-overridable server-side timeouts (statement_timeout=30s, lock_timeout=10s, idle_in_transaction_session_timeout=60s) applied as connection startup parameters, so a stalled transaction can't hold a per-slip FOR UPDATE lock indefinitely.

slippy/postgres_store_updates.go
smunukutla-mycarrier: Fixed in d469c3e: stepName is now validated (must be a configured step AND match ^[A-Za-z0-9_]+$) before being interpolated into the column identifier. An unknown/unsafe name skips the column write while still recording the component event (ClickHouse parity), which also resolves the 42703 event-rollback correctness bug you flagged.

postgres/postgres.go
smunukutla-mycarrier: Fixed in cde00d4: defaultSSLMode is now verify-full, with an optional POSTGRES_SSLROOTCERT override for a private CA / non-system trust store.

postgres/go.mod
smunukutla-mycarrier: Fixed in cde00d4 / 785389a: bumped golang.org/x/crypto to v0.54.0 in both modules to align with slippy.

postgresmigrator/migrator.go
smunukutla-mycarrier: Addressed in 785389a: documented on CreateTables the single-migrator (single pre-deploy Job) assumption, why a double-run is self-correcting (per-migration transaction + idempotent CREATE/ALTER ... IF [NOT] EXISTS), and the pg_advisory_lock path should concurrent migrators ever become possible.

slippy/postgres_store_updates.go
smunukutla-mycarrier: Fixed in d469c3e: the guard is now the ported ClickHouse freshness gate (window + push_parsed bypass + gate toggle), so step retries/re-runs are allowed once the prior terminal is older than the window. Integration tests cover failed→running after the window and the push_parsed re-push.

slippy/postgres_store_updates.go
smunukutla-mycarrier: Fixed in d469c3e: recomputeAggregate returns early on len(active)==0, so a component-less aggregate is no longer written as 'completed'. Test added.

slippy/postgres_store_updates.go
smunukutla-mycarrier: Done in d469c3e: extracted the logic into a single shared free function computeAggregateStatus (new aggregate_status.go); both stores now call it and pgComputeAggregateStatus / the ClickHouse method were removed.

slippy/client.go
```diff
@@ -175,11 +175,10 @@ func (c *Client) AbandonSlip(ctx context.Context, correlationID, supersededBy st
...
 
-	slip.Status = SlipStatusAbandoned
-
-	// With timestamp-based versioning, each update gets a unique nanosecond timestamp,
-	// so there are no version conflicts and Update always succeeds.
-	if err := c.store.Update(ctx, slip); err != nil {
+	// Abandon changes only the top-level status. Use the atomic status-only update rather
+	// than a Load+full-row Update, so a concurrent step/history write is not clobbered by
+	// this stale snapshot (last-write-wins). See SlipStore.UpdateSlipStatus.
+	if err := c.store.UpdateSlipStatus(ctx, correlationID, SlipStatusAbandoned); err != nil {
```
pkuzmenko-mycarrier, line 181: **Heads up: this quietly changes a live ClickHouse write path.** `AbandonSlip` used to load the slip, flip the status, and rewrite the whole row; now it fires a single atomic status-only update instead. That's almost certainly the better call — no stale-snapshot clobber — but it's a real behavior change on the path prod runs today, not the inert addition the PR description implies.

The reassuring part: `UpdateSlipStatus` itself is pre-existing and already has a real-ClickHouse integration test covering the abandoned status (`clickhouse_store_integration_test.go:1402-1424`, plus the negative case at `1445-1451`). Only the new wiring right here is mock-covered.

Worth a conscious nod from the reviewer, and ideally one client-level test that abandons a slip over the real store. Not a blocker.

postgres/postgres.go
```diff
@@ -0,0 +1,216 @@
...
+	DefaultMinConns        int32         = 2
+	DefaultConnMaxLifetime time.Duration = 60 * time.Second
+
+	// Server-side statement/lock/transaction timeouts, applied as connection
+	// startup parameters (see session.buildRuntimeParams). They cap the blast
+	// radius of a hung query or a transaction that acquires a per-slip FOR UPDATE
+	// lock and then stalls — without them one stuck transaction blocks every
+	// mutator for that slip indefinitely. Conservative for a control-plane store
+	// whose operations are sub-second; override per env when a workload needs more.
+	DefaultStatementTimeout       time.Duration = 30 * time.Second
```
pkuzmenko-mycarrier, line 31: **A subtle mismatch worth pinning down before the migrator ships.** This connection caps `statement_timeout` at 30s, but the migrator hands each migration a 5-minute budget (`postgresmigrator/migrator.go:93`). If the migrator ever runs over a pool built by this package — the natural pairing, since `NewMigratorFromSession` takes a `PoolProvider` — Postgres will kill any statement that runs past 30s long before the migrator's own deadline, and the extra 4.5 minutes of budget quietly become fiction.

Nothing in this PR actually wires that pairing yet (`RunPostgresMigrations` accepts whatever pool the caller hands it), so this is a trap laid for the follow-up consumer, not a live bug today. The one statement that can realistically trip it is the non-`CONCURRENTLY` `CREATE INDEX` on `routing_slips` (`postgres_migrations.go:213-236`) once that table grows — the `ADD COLUMN ... DEFAULT 'pending'` is a constant default, so it's metadata-only on PG 11+ and safe.

Easy fix: reset `statement_timeout` to `0` (or the migration timeout) on the migrator's session, or document that migrations must run on an uncapped pool.

slippy/postgres_store.go
```diff
@@ -0,0 +1,429 @@
...
+	args = append(args, slip.CorrelationID)
+
+	query := fmt.Sprintf("UPDATE routing_slips SET %s WHERE correlation_id = $%d",
+		strings.Join(sets, ", "), n)
+
+	// Serialize this full-row overwrite against the per-slip FOR UPDATE lock the step
+	// mutators hold, so it can no longer interleave with a concurrent updateStepTx commit
+	// and silently clobber it. (Load-then-Update callers such as AbandonSlip/PromoteSlip
+	// still snapshot before this lock; where they change only status they should prefer the
+	// atomic UpdateSlipStatus rather than a full-row write.)
```
pkuzmenko-mycarrier, line 127: **Small race worth a linked issue before it disappears.** `Update` grabs the per-slip `FOR UPDATE` lock and then overwrites every column from the caller's snapshot — but callers like `PromoteSlip` load that snapshot *before* the lock exists, so a step/history write that commits in the gap gets silently overwritten (and the full-row write also skips the I5 freshness guard for those columns).

This is a known trade-off — it matches the old ClickHouse last-write-wins behavior, and the comment right here says as much — so it's not a regression. The only ask: the comment calls it "tracked as a follow-up" but names no ticket. Please attach the real issue ID, so the "do we persist `PromotedTo` / fix the read-modify-write" decision doesn't quietly evaporate.

bcarlock-mycarrier (approved):

postgres/postgres.go
smunukutla-mycarrier: Good catch — and it's live, not just a follow-up trap: slippy-migrator builds its pool via `postgres.NewPostgresSession` (main.go:33), so the 30s `statement_timeout` did apply to migrations. Fixed in 4f9854f: each migration/revert transaction now issues `SET LOCAL statement_timeout = 0` (and `lock_timeout = 0`), so a migration is bounded solely by the migrator's own context regardless of the pool's app-facing caps; the reset is tx-scoped so pooled connections are unaffected. Added an integration test asserting both read `0` inside a migration.

slippy/postgres_store.go
smunukutla-mycarrier: Done — filed **DEVOPS-202** ("PromoteSlip drops PromotedTo and still uses a full-row read-modify-write") capturing both the unpersisted `PromotedTo` and the deferred RMW/persistence decision, and referenced it in the code comment (28e0e2c).

slippy/client.go
smunukutla-mycarrier: Conscious nod: agreed this is an intentional behavior change on the live path — and the better one (no stale-snapshot clobber, no lost concurrent history append). `AbandonSlip` is now a thin delegation to the pre-existing `UpdateSlipStatus`, which already has real-ClickHouse coverage (clickhouse_store_integration_test.go:1402-1451); only the new client-level wiring is mock-covered. Happy to add a client-level abandon-over-real-store test before merge if you'd like it — leaving this thread open for your call.

pkuzmenko-mycarrier (approved):

PR #81, opened 2026-07-28
feat(teamsbot): reusable Bot Connector client for send-only Teams delivery (DEVOPS-137)

smunukutla-mycarrier:
## What & why

Adds a new reusable module **`teamsbot`** — a Go client for the Microsoft **Bot Connector REST API** for **send-only** proactive messaging to Microsoft Teams channels. This is the foundational library for [DEVOPS-137](https://linear.app/mycarrier/issue/DEVOPS-137/ci-publish-release-notes-to-teams-channel-after-prod-release) (CI posting release notes to a Teams channel). Microsoft is deprecating Incoming Webhooks, so delivery must go through the Bot Framework.

Sending a proactive message is just two REST calls (client-credentials token → `POST /v3/conversations`), so there's **no need for the Bot Framework SDK** (which has no Go flavour anyway). This module keeps the whole thing to `net/http` + `golang.org/x/oauth2`.

## What's in it

A single new sibling module `teamsbot/` (`github.com/MyCarrier-DevOps/goLibMyCarrier/teamsbot`):

| File | Responsibility |
|------|----------------|
| `config.go` | `Config` + `LoadConfig`/`LoadConfigFromViper` (viper idiom, mirrors `argocdclient`) |
| `auth.go` | client-credentials `oauth2.TokenSource` (cached), single-tenant token endpoint + `api.botframework.com/.default` scope |
| `client.go` | `Sender` interface, `Client`, functional options, `NewClient`/`NewClientFromEnv`, `SendToChannel` |
| `types.go` | generic `Activity`/`Attachment` (+ `AdaptiveCardActivity`/`TextActivity` helpers), `SendResult` |
| `apierror.go` | structured `*APIError` (status/code/message + `Retry-After`), mirrors `pollyapi` |
| `teamsbottest/` | `MockSender` for consumers' tests |

## Design decisions worth knowing

- **Card-agnostic by design.** The card travels as `json.RawMessage` — the library never knows any specific card layout, so it's reusable by any service. (Polly builds its own Adaptive Card and hands it in.)
- **Send-only.** No inbound messaging endpoint, no Connector→bot JWT validation. That machinery is deliberately out of scope (nothing here needs to *receive*).
- **Token fetch is bounded by the client's HTTP timeout** — `NewClient` injects the client's `*http.Client` into the oauth2 context (via `oauth2.HTTPClient`) so a hung Entra ID endpoint can't block indefinitely (this was a review fix; see below).

## Public API

```go
c, err := teamsbot.NewClientFromEnv() // TEAMS_BOT_APP_ID/SECRET/TENANT_ID (+ optional TEAMS_SERVICE_URL)
_, err = c.SendToChannel(ctx, "19:...@thread.tacv2", tenantID, teamsbot.AdaptiveCardActivity(cardJSON))
```

## Testing & quality

- Table-driven tests against `httptest`, token injected via `WithTokenSource` (no real network in tests).
- **Coverage 94.0%** (module-scoped; exceeds the repo's 75% gate — we self-imposed >85% per the ticket).
- `make lint` 0 issues, `make check-sec` clean (2 informational transitive advisories in `x/text`/`x/sys`, not reachable — flagged for a future repo-wide bump).

## Security

- Client secret only ever goes to the token source (TLS to Entra ID); never logged, never in an error string. Token held in memory only.
- No inbound surface → no JWT-inbound attack surface.
- `parseAPIError` reads only the Bot Connector response (not the token endpoint's), with a 4 KiB body cap.

## Reviewer guide — where to look

1. `client.go` `SendToChannel` — request construction (path, `channelData` shape, auth header) and 2xx/non-2xx handling.
2. `auth.go` — token endpoint/scope + the `oauth2.HTTPClient` timeout injection in `client.go NewClient`.
3. `apierror.go` — error envelope parsing + `Retry-After`.

The bulk of the diff is straightforward transcription + tests; the two spots with real judgment are the token-timeout wiring and the connector request body.

## Notes / follow-ups

- ✅ **Validated end to end.** A local smoke test delivered a card to a real Teams channel with exactly this payload — the `POST /v3/conversations` body works **without** a top-level `bot` ChannelAccount (the connector infers the bot from the bearer token). The earlier open question is resolved; no change needed.
- Consumed by polly-api at the pre-release tag `teamsbot v1.3.99-feat-devops-137-teamsbot.9`; bump consumers to the released version once this merges.


teamsbot/client.go
```diff
@@ -0,0 +1,161 @@
...
+	defer resp.Body.Close() //nolint:errcheck // best-effort close
+
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
+		return zero, parseAPIError(resp)
+	}
+
+	var result SendResult
+	if err := json.NewDecoder(resp.Body).Decode(&result); err != nil {
+		return zero, fmt.Errorf("teamsbot: decode response: %w", err)
+	}
```
bcarlock-mycarrier, line 202: **A 2xx with an empty body is reported as a failure, so a delivered message gets retried and posts a duplicate thread.**

`json.NewDecoder(resp.Body).Decode(&result)` returns `io.EOF` when the body is empty, so a success status with no payload surfaces as an error:

```
204 No Content -> res={ConversationID: ActivityID:}  err=teamsbot: decode response: EOF
```

This matters more than it looks, because of how the consumer reacts. On the `polly-api` branch that consumes this module (`feat/devops-137-teams-delivery`, pinned at `teamsbot v1.3.99-feat-devops-137-teamsbot.9`), the send is wrapped in a retry loop:

```go
// internal/provider/teams/message.go
err = retry.Do(ctx, 0, func() (int, error) {
    _, sendErr := p.sender.SendToChannel(ctx, channelID, p.tenantID, activity)
    if sendErr == nil { return http.StatusOK, nil }
    var apiErr *tb.APIError
    if errors.As(sendErr, &apiErr) { return apiErr.Status, ... }
    return -1, &APIError{Status: http.StatusBadGateway, ...}   // <- EOF lands here
})
```

`teamsbot: decode response: EOF` is **not** an `*APIError`, so `errors.As` fails, the callback returns `-1`, and `shouldRetry(-1)` returns `true` (`internal/retry/retry.go:44-46`, `maxAttempts = 3`, `baseDelay = 3s`). So a delivered message is retried up to **3 times** — and because `POST /v3/conversations` starts a new reply chain each time, that is up to **three duplicate release-note cards in the channel**, with polly still reporting failure. For an at-least-once CI notifier, a false negative on the last step of a *successful* delivery is the expensive direction.

The same lines have a second problem: the success path has no size cap, unlike the 4 KiB-capped error path. Verified — a server returning ~8 MiB had the whole body allocated (`err=<nil>`, decoded length 8388608). `SendResult` is two short ID strings, so a 1 MiB ceiling is generous.

No `suggestion` block, because the fix needs `errors` and `io` added to the import block at lines 3-13 plus a new const. Verified compiling, `gofmt`/`vet` clean, suite green under `-race`:

```go
// add "errors" and "io" to the imports; next to defaultTimeout:
// maxResponseBody caps the success-path read. SendResult is two short IDs.
const maxResponseBody = 1 << 20

// ErrBadResponse marks a 2xx whose body could not be used. The message text is
// deliberately "teamsbot: decode response" so existing assertions keep matching.
var ErrBadResponse = errors.New("teamsbot: decode response")

	var result SendResult
	if err := json.NewDecoder(io.LimitReader(resp.Body, maxResponseBody)).Decode(&result); err != nil {
		switch {
		case errors.Is(err, io.EOF):
			// 200/204 with an empty body: the status already said the activity
			// was accepted, so there is nothing to decode and nothing failed.
			return result, nil
		case errors.Is(err, io.ErrUnexpectedEOF):
			return zero, fmt.Errorf("%w: body exceeds %d bytes", ErrBadResponse, maxResponseBody)
		default:
			return zero, fmt.Errorf("%w: %v", ErrBadResponse, err)
		}
	}
	return result, nil
```

Two notes on the choices. `io.LimitReader` rather than `http.MaxBytesReader`: the latter produces the message `http: request body too large`, which would permanently misreport a **response** as a request — the stdlib comment at `net/http/request.go:1198-1200` says that text cannot be changed. And exporting `ErrBadResponse` gives the consumer an `errors.Is` hook so it can stop retrying a send that may well have been delivered; without it, an oversized or malformed body is retried three times just like the empty one.

teamsbot/config.go
```diff
@@ -0,0 +1,82 @@
...
+		return fmt.Errorf("TEAMS_BOT_APP_ID is required")
+	}
+	if config.AppSecret == "" {
+		return fmt.Errorf("TEAMS_BOT_APP_SECRET is required")
+	}
+	if config.TenantID == "" {
+		return fmt.Errorf("TEAMS_BOT_TENANT_ID is required")
+	}
+	return nil
+}
```
bcarlock-mycarrier, line 121: **`TEAMS_SERVICE_URL` is never validated, so a query string silently misroutes the request and a plain-HTTP host transmits the bearer token in clear.**

`validateConfig` checks `AppID`, `AppSecret` and `TenantID` but not `ServiceURL` — even though it is the only field used to build a URL, by concatenation at `client.go:137` (`c.serviceURL + "/v3/conversations"`), with the Entra-issued token attached at `client.go:144`. Two misconfigurations verified against this HEAD:

```
TEAMS_SERVICE_URL=https://host/teams?trace=1
  -> server receives "/teams?trace=1/v3/conversations"
     err = <nil>          <-- success reported for a request that never hit the endpoint

TEAMS_SERVICE_URL=smba.trafficmanager.net/teams        (scheme omitted)
  -> teamsbot: send request: Post "smba.trafficmanager.net/teams/v3/conversations":
     unsupported protocol scheme ""        ... at first send, not at config load
```

The first is the serious one: the path lands inside the query string, so the library returns success for a request that never reached `/v3/conversations`. A release-notes job would report a delivery that never happened.

Separately, nothing constrains the scheme, so a plain-`http://` value transmits the token in cleartext (verified: `Authorization_received="Bearer ..."` over a non-TLS hop). Microsoft is explicit about this — from [Authenticate requests with the Bot Connector API](https://learn.microsoft.com/en-us/azure/bot-service/rest-api/bot-framework-rest-connector-authentication): *"Only specify the JWT token in the `Authorization` header of requests you send to the Bot Connector service. Do NOT send the token over unsecured channels... The JWT token... is like a password... Anyone that possesses the token may use it to perform operations on behalf of your bot."* The realistic trigger is not an attacker but an operator typo or a bad value threaded from a workflow template. The default (`https://smba.trafficmanager.net/teams`, `config.go:12`) is correct and safe.

No `suggestion` block — the fix needs a `net/url` import and touches two functions. Verified compiling with **zero test regressions**, `gofmt`/`vet` clean, `golangci-lint` 0 issues:

```go
// ValidateServiceURLShape rejects values that cannot address the Bot Connector.
// Kept separate from the https check so NewClient can enforce shape on any
// caller-built Config while the scheme rule stays on the env path.
func ValidateServiceURLShape(raw string) error {
	u, err := url.Parse(raw)
	if err != nil {
		return fmt.Errorf("teamsbot: TEAMS_SERVICE_URL is not a valid URL: %w", err)
	}
	if !u.IsAbs() || u.Host == "" {
		return fmt.Errorf("teamsbot: TEAMS_SERVICE_URL must be an absolute URL with a host, got %q", raw)
	}
	if u.RawQuery != "" || u.Fragment != "" {
		return fmt.Errorf("teamsbot: TEAMS_SERVICE_URL must not carry a query or fragment, got %q", raw)
	}
	return nil
}
```

`validateConfig` then calls `ValidateServiceURLShape` **and** additionally requires `https`; `NewClient` calls only `ValidateServiceURLShape`, recording any error on the `Client` for `SendToChannel` to surface — so no exported signature changes and the pinned polly consumer is unaffected.

Why the split, rather than enforcing `https` everywhere: enforcing it inside `NewClient` **breaks six existing tests**, because every send test points at `httptest.NewServer` (plain `http://127.0.0.1:...`). The split keeps the whole suite green while still closing the misroute on every path — verified with the server never reached (`RequestURI=""`).

Two things deliberately **not** done. First, no host allowlist: I initially proposed one, but it would reject three of Microsoft's four documented endpoints — the sovereign clouds are on different *hosts*, not different paths (GCC `smba.infra.gcc.teams.microsoft.com`, GCC High `smba.infra.gov.teams.microsoft.us`, DoD `smba.infra.dod.teams.microsoft.us`). Second, being straight about the residual gap: this fix stops cleartext and misroute, but a typo'd *https* host is still accepted. That is now a documentation control — see the README comment on this PR.

This also subsumes a related gap: `NewClient`'s doc says it "builds a Client from a validated Config", but nothing validates — `validateConfig` is reachable only from `LoadConfigFromViper` (`config.go:65`). Wiring the shape check into `NewClient` closes both.

teamsbot/config.go
```diff
@@ -0,0 +1,82 @@
...
+		"tenant_id":   "TEAMS_BOT_TENANT_ID",
+		"service_url": "TEAMS_SERVICE_URL",
+	}
+	for key, env := range binds {
+		if err := vp.BindEnv(key, env); err != nil {
+			return nil, fmt.Errorf("error binding %s: %w", env, err)
+		}
+	}
+	vp.AutomaticEnv()
+	return LoadConfigFromViper(vp)
```
bcarlock-mycarrier: **`vp.AutomaticEnv()` with no prefix lets bare `APP_ID` / `APP_SECRET` / `TENANT_ID` / `SERVICE_URL` silently override the documented `TEAMS_BOT_*` variables.**

All four keys are already explicitly bound at lines 42-46, then `AutomaticEnv()` is called with no `SetEnvPrefix`. In viper's `find()` (`viper@v1.21.0/viper.go:1226-1245`) the `automaticEnvApplied` block at 1227-1237 is consulted **before** `envkeys, exists := v.env[lcaseKey]` at 1238, and `mergeWithEnvPrefix` (418-424) returns `strings.ToUpper(in)` unchanged when the prefix is empty. So the generic names win.

The concrete, demonstrated harm is that the test suite is not hermetic. From this worktree at HEAD:

```
$ APP_ID=ambient-hijack SERVICE_URL=https://ambient.example go test ./...
--- FAIL: TestLoadConfigFromEnv (0.00s)
    config_test.go:84: got &{AppID:ambient-hijack ... ServiceURL:https://ambient.example}
FAIL
```

With the fix below, the same command passes. Any developer or runner with one of those four very generic names in the ambient environment gets a red build in a module they did not touch. Beyond CI flakiness, any deployment that happens to define one of them loses the documented configuration silently rather than loudly — and there is a specific consequence for the intended consumer: `polly-api/main.go` reads `TEAMS_BOT_TENANT_ID` directly with `os.Getenv` for the *wire* tenant but gets its *token* tenant from `teamsbot.LoadConfig()`. With a bare `TENANT_ID` present, teamsbot would mint a token for the hijacked tenant while polly puts the correct one on the wire.

`AutomaticEnv()` buys nothing here — all four keys are already explicitly bound, so deleting it preserves the documented behaviour exactly. Re-verified: `gofmt`/`vet` clean, `TestLoadConfigFromEnv` / `TestLoadConfigFromViper` / `TestNewClientFromEnv*` all pass.

**Please don't reach for `vp.SetEnvPrefix("TEAMS_BOT")` instead** — that is the natural-looking fix and it is wrong. `mergeWithEnvPrefix("service_url")` would then look up `TEAMS_BOT_SERVICE_URL`, silently diverging from the documented `TEAMS_SERVICE_URL` (`config.go:40`, `README.md:26`). Deletion is the only correct fix.

```suggestion
	}
	return LoadConfigFromViper(vp)
```

teamsbot/auth_test.go
```diff
@@ -0,0 +1,30 @@
...
+
+func TestNewTokenSource(t *testing.T) {
+	cfg := &Config{AppID: "id", AppSecret: "secret", TenantID: "tenant", ServiceURL: defaultServiceURL}
+	ts := newTokenSource(context.Background(), cfg)
+	if ts == nil {
+		t.Fatal("newTokenSource returned nil")
+	}
+	// Token() would make a network call, so we only assert construction here;
+	// end-to-end token use is covered in client_test via WithTokenSource.
+}
```
bcarlock-mycarrier, line 30: **The suite survives three independent mutations: the token exchange, `WithHTTPClient`, and the card payload are all unpinned.**

`auth.go` is one of the two files flagged in the PR description as needing real judgment, and it has no behavioral coverage. The three tests here cannot fail:

- `TestBotFrameworkScope` (16-20) compares a constant to its own literal — no independent oracle.
- `TestNewTokenSource` (22-30) asserts `ts != nil`, but `clientcredentials.Config.TokenSource` returns `oauth2.ReuseTokenSource(nil, src)` (`oauth2@v0.36.0/clientcredentials.go:79-85`), which is never nil.
- `TestTokenURL` checks the string template only.

That was confirmed by mutation rather than argued. Three changes that the suite does **not** detect — all 31 tests still pass after each:

1. Flipping `AuthStyleInParams` → `AuthStyleInHeader` **and** deleting the `oauth2.HTTPClient` injection in `NewClient` — i.e. both mechanisms the PR description highlights as the careful parts.
2. Reducing `WithHTTPClient` to a total no-op (`return func(c *Client) { _ = h }`).
3. `TestSendToChannelSuccess` (`client_test.go:60-71`) asserts `isGroup` and the `channelData` shape but never inspects `body["activity"]` — so nothing pins that the Adaptive Card reaches the wire with `contentType: application/vnd.microsoft.card.adaptive`, which is the one thing this module exists to do.

Coverage reports `newTokenSource 100.0%` and `tokenURL 100.0%`, which is exactly the "lines executed, behaviour unpinned" gap. To be clear: the implementation is *correct* — I verified the request matches Microsoft's documented single-tenant form byte for byte. The point is that nothing in the suite would have caught it if it weren't.

The replacement below turns that verification into a real test using a recording `RoundTripper`, which needs no new seam (`WithHTTPClient` already lets you swap the transport the token fetch travels on) and kills mutations 1 and 2. Re-verified: `gofmt` clean, `vet` clean, passes under `-race`, `golangci-lint` 0 issues.

Two follow-ups worth doing in the same change, neither of which needs a new file:

- **Fix `TestWithHTTPClientOverride`** (`client_test.go:158-182`) to assert on transport identity rather than on a successful send: `if !rt.used { t.Error("WithHTTPClient was ignored") }`, and for the nil case `if c2.httpClient.Transport != nil {...}`. Note **not** to assert on `Timeout` there — the redirect/timeout hardening suggested elsewhere on this PR normalizes every injected client's `Timeout` to `defaultTimeout`, which would make a timeout-based assertion tautological again.
- **Cover `config.go:75-77`.** The `TEAMS_BOT_APP_SECRET is required` branch has execution count `0` in the HEAD profile — it is the only unexercised block in `validateConfig` (hence its 85.7%). The table at `config_test.go:36-45` covers missing `app_id` and missing `tenant_id`, and `TestNewClientFromEnvMissingEnv` fails on `AppID` first. `AppSecret` is the one required field sourced from a secret manager rather than a plain env var, so it is the likeliest real misconfiguration and the one nothing tests. Two lines.

One heads-up: applying this PR's suggested fixes adds validation and error-handling branches that no test exercises yet, so module coverage moves from 93.8% to roughly 89% (I measured 89.1% with the suggestion blocks applied). Still far above the CI gate of 75 (`ci.yml:80`), but worth covering the new branches rather than letting it drift.

```suggestion
package teamsbot

import (
	"io"
	"net/http"
	"net/url"
	"strings"
	"testing"
)

// tokenRT answers the Entra ID token endpoint and records the form it was sent,
// so the client-credentials exchange can be asserted without a network call.
type tokenRT struct {
	url  string
	form url.Values
}

func (r *tokenRT) RoundTrip(req *http.Request) (*http.Response, error) {
	body, _ := io.ReadAll(req.Body)
	r.url = req.URL.String()
	r.form, _ = url.ParseQuery(string(body))
	return &http.Response{
		StatusCode: http.StatusOK,
		Header:     http.Header{"Content-Type": []string{"application/json"}},
		Body: io.NopCloser(strings.NewReader(
			`{"access_token":"AT","token_type":"Bearer","expires_in":3599}`)),
	}, nil
}

// TestTokenSourceExchange pins the single-tenant client-credentials request the
// Bot Connector requires: tenant-scoped endpoint, client_credentials grant, and
// the api.botframework.com/.default scope, with the secret in the form body.
// See https://learn.microsoft.com/azure/bot-service/rest-api/bot-framework-rest-connector-authentication?tabs=singletenant
func TestTokenSourceExchange(t *testing.T) {
	rt := &tokenRT{}
	cfg := &Config{AppID: "the-app", AppSecret: "the-secret", TenantID: "the-tenant"}
	// NewClient injects this http.Client into the oauth2 context, so the token
	// fetch is observable here -- which also proves the injection is wired up.
	c := NewClient(cfg, WithHTTPClient(&http.Client{Transport: rt}))

	if _, err := c.tokenSource.Token(); err != nil {
		t.Fatalf("Token(): %v", err)
	}

	wantURL := "https://login.microsoftonline.com/the-tenant/oauth2/v2.0/token"
	if rt.url != wantURL {
		t.Errorf("token URL = %q, want %q", rt.url, wantURL)
	}
	for key, want := range map[string]string{
		"grant_type":    "client_credentials",
		"client_id":     "the-app",
		"client_secret": "the-secret",
		"scope":         "https://api.botframework.com/.default",
	} {
		if got := rt.form.Get(key); got != want {
			t.Errorf("form[%s] = %q, want %q", key, got, want)
		}
	}
}
```

teamsbot/go.mod
```diff
@@ -0,0 +1,25 @@
...
+	github.com/pelletier/go-toml/v2 v2.2.4 // indirect
+	github.com/sagikazarmark/locafero v0.11.0 // indirect
+	github.com/sourcegraph/conc v0.3.1-0.20240121214520-5f936abd7ae8 // indirect
+	github.com/spf13/afero v1.15.0 // indirect
+	github.com/spf13/cast v1.10.0 // indirect
+	github.com/spf13/pflag v1.0.10 // indirect
+	github.com/subosito/gotenv v1.6.0 // indirect
+	go.yaml.in/yaml/v3 v3.0.4 // indirect
+	golang.org/x/sys v0.29.0 // indirect
+	golang.org/x/text v0.28.0 // indirect
```
bcarlock-mycarrier: **This module's transitive set is systematically older than every sibling, and it reintroduces a HIGH advisory the rest of the repo has already shed.**

| Dependency | teamsbot | every sibling module |
|---|---|---|
| `golang.org/x/text` | **v0.28.0** | v0.40.0 |
| `golang.org/x/sys` | **v0.29.0** | v0.47.0 |
| `github.com/sagikazarmark/locafero` | **v0.11.0** | v0.12.0 |
| `github.com/sourcegraph/conc` | **present** (pseudo-version) | absent |

Confirmed against OSV by package and version:

- `golang.org/x/text@v0.28.0` — **GO-2026-5970 / CVE-2026-56852**, "Infinite loop on invalid input in golang.org/x/text", fixed in **v0.39.0**. Trivy rates it **HIGH**. `x/text@v0.40.0` (the repo standard) returns clean.
- `golang.org/x/sys@v0.29.0` — **GO-2026-5024 / CVE-2026-39824**, integer overflow in `NewNTUnicodeString`, fixed in v0.44.0. Windows-only, so genuinely informational for Linux CI.

**Your reachability claim is correct and I want to be explicit about that:** `govulncheck` places both under Package/Module results rather than Symbol results, and `make check-sec PKG=teamsbot` passes with "0 vulnerabilities". Neither is called by this module.

Where the PR description is misleading is in calling these "informational" and deferring them to "a future repo-wide bump". There is nothing repo-wide left to do — every sibling module is *already* on the fixed versions, so this is a teamsbot-only regression, fixed by one command. It still matters despite being unreachable here because `teamsbot` is a **published** module: a consumer that imports only `teamsbot` inherits `x/text v0.28.0` through MVS and may reach the vulnerable code through its own use of the package. I verified that concretely — a module requiring only `teamsbot` and importing `golang.org/x/text/unicode/norm` gets the advisory under **Symbol Results** with a called trace. A library shouldn't lower the floor for its consumers.

In fairness, the counter-case bounds the impact: a consumer importing `teamsbot` *and* any other goLibMyCarrier module resolves to v0.40.0 / v0.47.0 by MVS anyway, so the floor only bites teamsbot-only consumers. That doesn't change the ask, since the fix is one command.

The `locafero`/`conc` row has the same root cause: `locafero v0.12.0` dropped `sourcegraph/conc`, so pinning v0.11.0 uniquely drags a pseudo-version (`v0.3.1-0.20240121214520-5f936abd7ae8`, an unreleased Jan-2024 commit) into the graph that no sibling carries. `go mod verify` reports all modules verified, so this is stale pinning, not tampering — it looks like the graph was resolved against a stale module cache rather than tidied against the repo's current floor.

**Deliberately not offering a `suggestion` block here.** Editing just these two lines is what GitHub's "Commit suggestion" button would do, and it breaks the build:

```
go build ./... ->
  fsnotify@v1.9.0/internal/debug_linux.go:9:2: missing go.sum entry for golang.org/x/sys/unix
  afero@v1.15.0/util.go:28:2: missing go.sum entry for golang.org/x/text/runes
```

`go mod download` exits 0 without repairing it, so the `unit-test-teamsbot` job would fail. Please run the command instead — verified end to end (all three upgraded, `conc` dropped, tests green, "0 vulnerabilities in modules you require"):

```
go get golang.org/x/text@v0.40.0 golang.org/x/sys@v0.47.0        github.com/sagikazarmark/locafero@v0.12.0 && go mod tidy
```

teamsbot/client.go
```diff
@@ -0,0 +1,161 @@
...
+		opt(c)
+	}
+	if c.tokenSource == nil { // WithTokenSource may have set it
+		if cfg == nil {
+			cfg = &Config{}
+		}
+		ctx := context.WithValue(context.Background(), oauth2.HTTPClient, c.httpClient)
+		c.tokenSource = newTokenSource(ctx, cfg)
+	}
+	return c
```
bcarlock-mycarrier, line 116: **Harden the effective HTTP client once: bound a missing timeout and refuse redirects, for the default and injected clients alike.**

Two defects meet on these lines, and both are cheapest to fix in one place — after the option loop, before line 87 captures `c.httpClient` for the token source.

**1. A redirect can carry the bearer token to another port, a sibling subdomain, or a downgraded `http://` URL.** No `CheckRedirect` is set, so the stdlib default follows up to 10 redirects. The good news first: Go *does* strip a manually-set `Authorization` header on a genuine cross-host redirect, so there is no cross-host exfiltration here (verified: `attacker_Authorization=""`). But `net/http`'s `shouldCopyHeaderOnRedirect` (`src/net/http/client.go:1005-1022`, delegating to `isDomainOrSubdomain` at 1028-1045) compares **hostnames only and never inspects the scheme**, and `idnaASCIIFromURL` uses `url.Hostname()`, which strips the port. Verified consequences:

```
same host, different port  -> Authorization="Bearer SUPER-SECRET-BOT-TOKEN"
https -> http, same host   -> final_hop_over_TLS=false  Authorization="Bearer SUPER-SECRET-BOT-TOKEN"
```

Refusing redirects costs nothing functionally and fails closed — `parseAPIError` turns the 3xx into an `*APIError` carrying the redirect status.

**2. `WithHTTPClient` accepts a zero-timeout client, silently unbounding both calls.** The PR description's timeout guarantee is real for the default path and I verified it, including the `oauth2.HTTPClient` injection. But `WithHTTPClient` takes the caller's client verbatim, and because options are applied *before* the token source is built, a zero-`Timeout` client removes the only bound on the token fetch — the caller's `ctx` never reaches it (`context.Background()` at line 87). Measured: with a 150 ms caller deadline and `WithHTTPClient(&http.Client{})`, a stalled Entra endpoint blocked for **3.002 s** and would block indefinitely. This isn't hypothetical — `client_test.go:168` passes exactly such a client, and it's the natural shape when wrapping a transport for OpenTelemetry (`&http.Client{Transport: otelhttp.NewTransport(...)}`).

The patch copies the client by value so we never mutate one the caller may share elsewhere. Re-verified standalone: range asserted byte-for-byte against HEAD, `gofmt` clean, `go build`/`go vet` clean, suite green under `-race`, `golangci-lint` 0 issues. Both leaks close:

```
https->http downgrade: err=teamsbot: bot connector 307  Authorization_on_cleartext_hop=""
same-host diff-port:   Authorization=""
caller's client Timeout after construction: 0s (not mutated)
```

Two caveats worth your judgement. This sets `CheckRedirect` **unconditionally**, which would override a redirect policy a caller deliberately supplied via `WithHTTPClient` — reasonable here since this endpoint has no legitimate redirect, but if you'd rather respect caller intent, gate it with `if hardened.CheckRedirect == nil`. And copying by value means a caller who later mutates `h.Timeout` no longer affects this client; that's almost certainly what you want, but it is a behaviour change worth knowing about.

```suggestion
	for _, opt := range opts {
		opt(c)
	}
	// Harden whatever client we ended up with -- the default above or one injected
	// via WithHTTPClient -- before the token source captures it below, so the token
	// fetch inherits the same bounds as the send. Copied by value so we never
	// mutate a client the caller may share; note a caller who later changes
	// h.Timeout no longer affects this client.
	//
	// Timeout: a zero value means "no timeout" and would let a CI job hang.
	// CheckRedirect: net/http's shouldCopyHeaderOnRedirect compares hostnames only
	// and never inspects the scheme, so a redirect to another port, a sibling
	// subdomain, or a downgraded http:// URL would carry the Authorization header.
	hardened := *c.httpClient
	if hardened.Timeout == 0 {
		hardened.Timeout = defaultTimeout
	}
	hardened.CheckRedirect = func(*http.Request, []*http.Request) error {
		return http.ErrUseLastResponse
	}
	c.httpClient = &hardened
	if c.tokenSource == nil { // WithTokenSource may have set it
		if cfg == nil {
			cfg = &Config{}
		}
		ctx := context.WithValue(context.Background(), oauth2.HTTPClient, c.httpClient)
		c.tokenSource = newTokenSource(ctx, cfg)
	}
	return c
```

teamsbot/config.go
```diff
@@ -0,0 +1,82 @@
...
+
+// Config holds the credentials and endpoint for the Teams bot. AppID, AppSecret,
+// and TenantID are required; ServiceURL is optional and defaults to the global
+// Teams Bot Connector URL.
+type Config struct {
+	AppID      string `mapstructure:"app_id"`
+	AppSecret  string `mapstructure:"app_secret"`
+	TenantID   string `mapstructure:"tenant_id"`
+	ServiceURL string `mapstructure:"service_url"`
+}
```
bcarlock-mycarrier, line 27: **`Config` discloses `AppSecret` under every common formatting verb and under `json.Marshal`.**

The PR description says the secret is "never logged, never in an error string", and I traced that and found it holds for the code *inside* this package — `AuthStyleInParams` puts the secret in the POST body rather than a URL, and `oauth2`'s `RetrieveError.Error()` echoes only the response, so wrapping it with `%w` can't surface it. The gap is the exported type itself, which has no `String()`/`GoString()` and no `json:"-"`:

```
%v  = &{app P@ssw0rd-LEAKED tid }
%+v = &{AppID:app AppSecret:P@ssw0rd-LEAKED TenantID:tid ServiceURL:}
%#v = &teamsbot.Config{AppID:"app", AppSecret:"P@ssw0rd-LEAKED", ...}
json.Marshal = {"AppID":"app","AppSecret":"P@ssw0rd-LEAKED","TenantID":"tid","ServiceURL":""}
```

"Never logged" is a property this package can't enforce on its consumers unless the type defends itself. One `log.Printf("config: %+v", cfg)` or one `slog` JSON attribute puts a live bot secret into CI logs. Note that `zap.Any` happens to be safe only by an accident of dispatch order (`case fmt.Stringer` before `default: Reflect` in `zap@v1.28.0/field.go:618-620`) — that's luck, not a defence, and `zap.Reflect` and `slog`'s JSONHandler were never covered. Entra's guidance is blunt: *"these credentials must be kept safe. Never publish that credential in your source code..."* ([client credentials flow](https://learn.microsoft.com/en-us/entra/identity-platform/v2-oauth2-client-creds-grant-flow)).

Since this module owns the type, this is the right place to fix, and it costs nothing — the struct is only ever mapstructure-decoded, so `json:"-"` removes no functionality. The value receiver is deliberate: it puts the method in the method set of both `Config` and `*Config`, so `%v` is covered either way.

Re-verified: `gofmt` clean, `vet` clean, suite green under `-race`, `golangci-lint` 0 issues, and all four disclosure paths closed (`AppSecret:REDACTED`, and `json.Marshal` no longer emits the field).

One formatting trap worth knowing: the repo's formatter wants **two spaces** between the struct tags on the `AppSecret` line. A single space fails `golangci-lint fmt` and would red-light CI. The block below has it right.

```suggestion
type Config struct {
	AppID      string `mapstructure:"app_id"`
	AppSecret  string `mapstructure:"app_secret"  json:"-"`
	TenantID   string `mapstructure:"tenant_id"`
	ServiceURL string `mapstructure:"service_url"`
}

// String implements fmt.Stringer so printing a Config never discloses AppSecret.
// The value receiver puts this in the method set of both Config and *Config.
func (c Config) String() string {
	return fmt.Sprintf("teamsbot.Config{AppID:%q, AppSecret:REDACTED, TenantID:%q, ServiceURL:%q}",
		c.AppID, c.TenantID, c.ServiceURL)
}

// GoString implements fmt.GoStringer so %#v also redacts AppSecret.
func (c Config) GoString() string { return c.String() }
```

teamsbot/apierror.go
```diff
@@ -0,0 +1,69 @@
...
+func parseAPIError(resp *http.Response) *APIError {
+	apiErr := &APIError{Status: resp.StatusCode}
+
+	raw, _ := io.ReadAll(io.LimitReader(resp.Body, maxErrorBody)) //nolint:errcheck // best-effort read
+
+	var w botError
+	if json.Unmarshal(raw, &w) == nil {
+		apiErr.Code = w.Error.Code
+		apiErr.Message = w.Error.Message
+	}
```
bcarlock-mycarrier, line 68: **The Teams-specific 403 body yields an empty `APIError.Code`, and that empty code reaches the consumer's own API response.**

`botError` models only the Connector's `ErrorResponse` envelope (`{"error":{"code","message"}}`), which is right for generic Connector errors. But Microsoft documents a second shape in [Send proactive messages](https://learn.microsoft.com/en-us/microsoftteams/platform/bots/how-to/conversations/send-proactive-messages):

```http
HTTP/1.1 403 Forbidden
{"errorCode":209,"message":"{\n  \"subCode\": \"MessageWritesBlocked\",\n  \"details\": \"Thread is blocked from message writes.\" ...}"}
```

Fed that body, the code at HEAD produces `Code = ""` — `json.Unmarshal` *succeeds* because unknown fields are ignored, so the fallback at line 60 dumps the raw JSON into `Message`. The docstring at lines 17-22 tells consumers to branch on the struct, but there is nothing machine-readable to branch on.

This is consumer-visible rather than academic: on the polly branch, `apiErr.Code` is copied verbatim into polly's own `APIError.Code` (`internal/provider/teams/message.go:51`) and returned to its HTTP clients, so the empty code propagates outward.

Being precise about severity, since it affects how you'd prioritise this: the value is **diagnostic, not control-flow**. Polly's `shouldRetry(403)` is already `false` (`internal/retry/retry.go:47-48`), so it stops retrying on any 403 by status alone and does not need `Code` to make that decision. And I'm dropping the stronger claim I first made that this is the error a Teams bot hits most often — the doc frames 209 per-user ("verify if a **user** has blocked or uninstalled"), whereas this module posts `isGroup: true` to a channel. What keeps it worth fixing is that `details` is thread-scoped ("Thread is blocked from message writes"), which is exactly what a moderated or archived channel produces, and losing the code means an operator debugging a silent release-notes failure gets an opaque blob instead of `209`.

Worth noting `pollyapi/apierror.go:32-45` deliberately used a superset struct for exactly this reason ("Decoding into one struct lets a single code path handle both") — that flexibility was dropped in the port.

The block below restores it. `strconv` and `strings` are already imported. Re-verified: `gofmt`/`vet` clean, 31 tests pass under `-race`, `golangci-lint` 0 issues, `Code == "209"` for the Teams shape, and the existing envelope / plain-text / empty-body cases all behave as before. The whitespace collapse is deliberate — without it the patch promotes the indented inner JSON into `Message`, giving a multi-line `error` string, and the consumer logs it with `logger.Errorf("...%v", err)`, which would break line-oriented ingestion.

```suggestion
// botError covers both error body shapes the Bot Connector emits:
//   - the documented ErrorResponse envelope {"error":{"code","message"}}
//   - the Teams shape {"errorCode":209,"message":"..."} used when a thread is
//     blocked from message writes
type botError struct {
	Error struct {
		Code    string `json:"code"`
		Message string `json:"message"`
	} `json:"error"`
	ErrorCode int    `json:"errorCode"`
	Message   string `json:"message"`
}

// parseAPIError builds an *APIError from a non-2xx response. It tolerates both
// JSON envelopes, plain-text bodies, and empty bodies, and recovers Retry-After
// from the header for 429s.
func parseAPIError(resp *http.Response) *APIError {
	apiErr := &APIError{Status: resp.StatusCode}

	raw, _ := io.ReadAll(io.LimitReader(resp.Body, maxErrorBody)) //nolint:errcheck // best-effort read

	var w botError
	if json.Unmarshal(raw, &w) == nil {
		apiErr.Code = w.Error.Code
		apiErr.Message = w.Error.Message
		if apiErr.Code == "" && w.ErrorCode != 0 {
			apiErr.Code = strconv.Itoa(w.ErrorCode)
		}
		if apiErr.Message == "" {
			// Collapse whitespace: the Teams shape nests an indented JSON blob in
			// this field, and an error string must stay on one log line.
			apiErr.Message = strings.Join(strings.Fields(w.Message), " ")
		}
	}
```

teamsbot/client.go
```diff
@@ -0,0 +1,161 @@
...
+// returns the resulting conversation and activity IDs. On any non-2xx response
+// it returns an *APIError.
+func (c *Client) SendToChannel(ctx context.Context, channelID, tenantID string, act Activity) (SendResult, error) {
+	var zero SendResult
+	if strings.TrimSpace(channelID) == "" {
+		return zero, fmt.Errorf("teamsbot: channelID is required")
+	}
+	if strings.TrimSpace(tenantID) == "" {
+		return zero, fmt.Errorf("teamsbot: tenantID is required")
+	}
```
bcarlock-mycarrier, line 143: **A nil or `null` card serialises to `"content":null` and leaves the library with no error.**

`channelID` and `tenantID` are guarded but the activity is not, so a caller can produce a body the Connector will reject with no local signal. Verified against HEAD:

```
AdaptiveCardActivity(nil)
  -> {..."attachments":[{"contentType":"application/vnd.microsoft.card.adaptive","content":null}]}
     err = <nil>
```

What makes this worth closing is the inconsistency: *invalid* card JSON **is** caught, with a good error —

```
AdaptiveCardActivity(json.RawMessage(`{not json`))
  -> teamsbot: marshal conversation: json: error calling MarshalJSON for type json.RawMessage: ...
```

— so a caller whose card builder returned `nil` on an error path (the easiest mistake to make with a `json.RawMessage`-based API) gets a round trip and an opaque connector 400, while a caller with a typo gets a clear local error.

The check tests the **value**, not just the length, because `json.Marshal(nil)` returns the 4-byte literal `null`: a length-only guard would pass it and still put `"content":null` on the wire. `bytes` is already imported at line 4. Re-verified: `gofmt`/`vet` clean, all 31 existing tests pass under `-race`, `golangci-lint` 0 issues; `nil`, `null` and `"  null\n"` are all rejected while a real card and `TextActivity` (no attachments) pass through.

Two notes. This does slightly narrow the promise at `types.go:16-17` that `Content` is "opaque JSON built by the caller — this module never inspects it"; checking for absence isn't really inspection, but it'd be worth rewording that to "never inspects its structure". And I deliberately left out an `act.Type == ""` guard — that can only fire for a hand-built `Activity{}`, since both constructors set it, which makes it a compile-visible programming error rather than something worth a runtime check.

```suggestion
	if strings.TrimSpace(channelID) == "" {
		return zero, fmt.Errorf("teamsbot: channelID is required")
	}
	if strings.TrimSpace(tenantID) == "" {
		return zero, fmt.Errorf("teamsbot: tenantID is required")
	}
	for i := range act.Attachments {
		// Reject absent content -- including the literal `null` that json.Marshal
		// returns for a nil value -- rather than posting an empty card.
		if c := bytes.TrimSpace(act.Attachments[i].Content); len(c) == 0 || string(c) == "null" {
			return zero, fmt.Errorf("teamsbot: attachment %d has no content", i)
		}
	}
```

teamsbot/README.md
```diff
@@ -0,0 +1,28 @@
...
+## Config
+
+| Env | Required | Default |
+|-----|----------|---------|
+| `TEAMS_BOT_APP_ID` | yes | — |
+| `TEAMS_BOT_APP_SECRET` | yes | — |
+| `TEAMS_BOT_TENANT_ID` | yes | — |
+| `TEAMS_SERVICE_URL` | no | `https://smba.trafficmanager.net/teams` |
+
+Testing: use `teamsbottest.MockSender`.
```
bcarlock-mycarrier, line 56: **Two constraints a consumer will hit are only discoverable by reading the source.**

The README is accurate as far as it goes — I checked every claim: the env var names match `config.go:36-41`, the default matches `config.go:12`, both signatures match, and the example would compile. Two things are missing:

1. **The bot registration must be single-tenant.** `tokenURL` (`auth.go:14-16`) hardcodes `https://login.microsoftonline.com/{tenantID}/oauth2/v2.0/token`. Per the [Connector auth doc](https://learn.microsoft.com/en-us/azure/bot-service/rest-api/bot-framework-rest-connector-authentication?tabs=singletenant), a **multi-tenant** bot must use the `botframework.com` authority instead. Someone who registers a multi-tenant Azure Bot and sets `TEAMS_BOT_TENANT_ID` to their own tenant gets an opaque Entra failure. The doc comment on `auth.go:13` says "single-tenant", but `Config.TenantID` and the README — the two places a consumer actually looks — don't. I'm not suggesting a multi-tenant knob; just document the constraint.

2. **Every `SendToChannel` creates a new conversation.** `POST /v3/conversations` starts a new reply chain, so N releases produce N separate channel threads rather than replies in one. `client.go:103-105` says so; the README doesn't, and for a release-notes poster this is the single most consequential behavioural fact. `SendResult` already carries the `id`/`activityId` needed to reply into or update the thread later — a consumer just has to know to keep them.

The sovereign-cloud URLs are worth naming too, since the table lists only the public default, and `TEAMS_SERVICE_URL` is the only escape hatch for them. This matters a bit more than usual given the validation comment elsewhere on this PR: documenting the legitimate hosts *is* the control against a typo'd host, since the library deliberately doesn't allowlist them. Note Microsoft's own caveat — *"These URLs are for proactive messages only. Avoid hardcoding them. Instead, use `serviceUrl` from the incoming activity"* — which this module cannot follow, since it is send-only and has no inbound activity to read a `serviceUrl` from; the config value is the only option available.

Verified the block applies cleanly at these lines and preserves the trailing "Testing:" line.

```suggestion
## Config

| Env | Required | Default |
|-----|----------|---------|
| `TEAMS_BOT_APP_ID` | yes | — |
| `TEAMS_BOT_APP_SECRET` | yes | — |
| `TEAMS_BOT_TENANT_ID` | yes | — |
| `TEAMS_SERVICE_URL` | no | `https://smba.trafficmanager.net/teams` |

The bot registration must be **single-tenant**: tokens are acquired from
`https://login.microsoftonline.com/{TEAMS_BOT_TENANT_ID}/oauth2/v2.0/token`. A
multi-tenant bot needs the `botframework.com` authority instead and is not supported.

Set `TEAMS_SERVICE_URL` for sovereign clouds — GCC
`https://smba.infra.gcc.teams.microsoft.com/teams`, GCC High
`https://smba.infra.gov.teams.microsoft.us/teams`, DoD
`https://smba.infra.dod.teams.microsoft.us/teams`.

Each `SendToChannel` call creates a **new conversation thread** in the channel; it does
not reply to an existing one. Keep the returned `ConversationID`/`ActivityID` if you
later want to reply into or update the message.

Testing: use `teamsbottest.MockSender`.
```

teamsbot/types.go
```diff
@@ -0,0 +1,65 @@
...
+			ContentType: AdaptiveCardContentType,
+			Content:     card,
+		}},
+	}
+}
+
+// TextActivity builds a plain-text "message" Activity.
+func TextActivity(text string) Activity {
+	return Activity{Type: "message", Text: text}
+}
```
bcarlock-mycarrier, line 38: **`TextActivity` is exported but has no consumer.**

`AdaptiveCardActivity` has a real one — on the `polly-api` branch that consumes this module (`feat/devops-137-teams-delivery`), `internal/provider/teams/message.go:39` calls `tb.AdaptiveCardActivity(card)`. I swept all 55 Go files on that branch for `TextActivity` and found **zero** occurrences. Within this module it appears only in `client_test.go` (7 call sites) and `types_test.go:32-37`.

The house rule in `CLAUDE.md` is explicit — "no extension points without a current consumer", and "when tempted to add an abstraction, count concrete call sites first" — and every exported symbol here becomes a compatibility commitment for every downstream consumer once the shared release tag lands.

To argue the other side honestly: it's three lines, `Activity` is fully exported so callers can write `Activity{Type: "message", Text: t}` themselves, and a plain-text helper is unsurprising in a Teams-sending library. This is the softest finding in the review and a one-word "keeping it deliberately" settles it.

No `suggestion` block, because deleting these lines breaks the two test files above — if you do drop it, the tests should move to `Activity{Type: "message", Text: "..."}` in the same change.

For contrast, and to be clear this isn't a blanket objection to exported surface: I specifically did **not** flag the `Sender` interface, even though a one-method interface declared next to its only implementation usually is a dependency-inversion smell. It has a genuine consumer-side use — polly's `provider.go:19,30` takes `sender tb.Sender` as an injected constructor parameter — which is exactly the port pattern the repo's DI guidance asks for.

bcarlock-mycarrier (changes requested):

teamsbot/client.go
smunukutla-mycarrier: Fixed in 6f56b0a — an empty 2xx now returns success: the decode uses `io.LimitReader` and treats `io.EOF` as an empty-but-accepted body, so a delivered message is no longer reported as failed (and no longer retried into duplicate threads). The success read is also capped (`maxResponseBody = 1<<20`) with an exported `ErrBadResponse`. Tests cover empty-body→success and oversized→ErrBadResponse.

teamsbot/config.go
smunukutla-mycarrier: Fixed in 3d8b462 — added `ValidateServiceURLShape` (rejects non-absolute, missing-host, query, and fragment) plus an `https`-scheme requirement on the LoadConfig/env path. `NewClient` enforces the shape check and surfaces any error from `SendToChannel`, so a query-string misroute or plain-http token leak is caught. Tests added.

teamsbot/config.go
smunukutla-mycarrier: Fixed in 3d8b462 — `LoadConfig` now reads via `os.Getenv` and no longer calls `AutomaticEnv()`, so bare `APP_ID`/`APP_SECRET`/`TENANT_ID`/`SERVICE_URL` can no longer shadow the `TEAMS_BOT_*` vars. `LoadConfigFromViper` is retained as the opt-in secret-manager seam.

teamsbot/auth_test.go
smunukutla-mycarrier: Fixed in 3d8b462 — added `TestTokenSourceExchange` using a recording `RoundTripper` (pins the single-tenant token URL, `grant_type`, `client_id`, `client_secret`, and the `api.botframework.com/.default` scope), a transport-identity assertion on `WithHTTPClient`, coverage of the missing-`AppSecret` branch, and an assertion that the Adaptive Card `contentType` reaches the wire. Mutation-verified as load-bearing.

teamsbot/go.mod
smunukutla-mycarrier: Fixed in 3d8b462 — bumped `x/text`→v0.40.0, `x/sys`→v0.47.0, `locafero`→v0.12.0 (dropped `sourcegraph/conc`), matching the sibling modules. `make check-sec PKG=teamsbot` is clean; the x/text HIGH advisory (GO-2026-5970) is cleared.

teamsbot/client.go
smunukutla-mycarrier: Fixed in 6f56b0a — `NewClient` now copies the effective client by value, bounds a zero `Timeout` to `defaultTimeout`, and sets `CheckRedirect` to refuse redirects (`http.ErrUseLastResponse`), so the bearer token cannot be carried across a port/scheme redirect and the token fetch inherits the timeout. A test asserts a 307 is not followed.

teamsbot/config.go
smunukutla-mycarrier: Fixed in 3d8b462 — `AppSecret` is tagged `json:"-"` and `Config` has value-receiver `String()`/`GoString()` that redact it. A test asserts `%v`/`%+v`/`%#v` and `json.Marshal` never contain the secret.

teamsbot/apierror.go
smunukutla-mycarrier: Fixed in 6f56b0a — `botError` now also parses the Teams `{"errorCode":209,"message":...}` shape; `parseAPIError` populates `Code` from `errorCode` when the envelope code is empty and collapses whitespace in the message. Test asserts `Code=="209"`; existing envelope/plain/empty cases unchanged.

teamsbot/client.go
smunukutla-mycarrier: Fixed in 6f56b0a — `SendToChannel` now rejects any attachment whose `Content` is empty or the literal `null`, so a nil card fails locally instead of posting `"content":null`. Test added.

teamsbot/README.md
smunukutla-mycarrier: Fixed in 3d8b462 — the README now documents the single-tenant requirement (multi-tenant needs the botframework.com authority and is unsupported), that each `SendToChannel` creates a new conversation thread (keep the returned ids to reply/update), and the sovereign-cloud `TEAMS_SERVICE_URL` values.

teamsbot/types.go
smunukutla-mycarrier: Keeping `TextActivity` deliberately — it is a small symmetric helper next to `AdaptiveCardActivity` and a natural part of a Teams-sending API (`Activity` is fully exported, so it is a thin convenience, not an abstraction). Happy to drop it if you would rather; leaving it for now.

bcarlock-mycarrier (approved):

